Issue #109, opened 2023-09-30
Restructure config YAML files to accommodate multiple species


leewujung:
To expand the package to handle krill inversion, we need to restructure the config YAML files. Below are the specifics, which we'll continue to add to in the next few weeks:

### Tasks
- [ ] determine if to keep the current 2 config YAML, one for data sources and one for processing parameters
  - my (@leewujung) vote is yes to keep this separation
- [ ] upgrade the YAML, in particular the data sources one, to use a tree-like structure, so that different categories of inputs are grouped naturally
- [ ] determine where the hake vs krill indicator should go in the above grouping
  - I think we should make it extensible for future expansion to other species

leewujung:
Sounds good.

PR #121, opened 2023-11-29
Use hierarchical structure in yaml file to specifiy data files

emiliom:
Re-organized the elements that define the input data files in `survey_year_2019_config.yml` to use a hierarchical structure that clearly groups elements by file type.

Partially addresses #109, and builds on the structure drafted in https://github.com/uw-echospace/EchoPro/issues/111#issuecomment-1751866907

### DIFF
--- a/config_files/survey_year_2019_config.yml
+++ b/config_files/survey_year_2019_config.yml
@@ -4,94 +4,81 @@
 # with the data_root_dir path also set below.
 
 ---
-  ##############################################################################
-  # Parameters
-  #########################################
-  survey_year: 2019            # survey year being considered
-  species_id: 22500            # target species ID for the survey year
-  CAN_haul_offset: 200         # The value to be added to the Canadian's haul number
+##############################################################################
+# Parameters
 
-  hemisphere: NW               # north/south and west/east hemispheres, used for the reports
-  ##############################################################################
+survey_year: 2019            # survey year being considered
+species_id: 22500            # target species ID for the survey year
+CAN_haul_offset: 200         # The value to be added to the Canadian's haul number
+hemisphere: NW               # north/south and west/east hemispheres, used for the reports
 
+##############################################################################
+# Directory path that contains all input data needed
 
-  # directory path that contains all data needed
-  data_root_dir: ../../2019_consolidated_files/
+data_root_dir: /usr/mayorgadat/workmain/acoustics/2021-NWFSC-EchoPro-HakeIGP/EchoPro/EchoProPython/2019_consolidated_files
+#  data_root_dir: ../../2019_consolidated_files/
 
-  ##############################################################################
-  # Paths to biological files #
-  #############################
+##############################################################################
+# Input data files
 
-  # File names and associated sheet name for the US section
-  length_US_filename: Biological/US/2019_biodata_length.xlsx
-  length_US_sheet: biodata_length
-  specimen_US_filename: Biological/US/2019_biodata_specimen_AGES.xlsx
-  specimen_US_sheet: biodata_specimen
-  catch_US_filename: Biological/US/2019_biodata_catch.xlsx
-  catch_US_sheet: biodata_catch
-
-
-  # File names for the Canada section
-  length_CAN_filename: Biological/CAN/2019_biodata_length_CAN.xlsx
-  length_CAN_sheet: biodata_length_CAN
-  specimen_CAN_filename: Biological/CAN/2019_biodata_specimen_CAN_AGES.xlsx
-  specimen_CAN_sheet: biodata_specimen_CAN
-  catch_CAN_filename: Biological/CAN/2019_biodata_catch_CAN.xlsx
-  catch_CAN_sheet: biodata_catch_CAN
-  ##############################################################################
-
-
-  ##############################################################################
-  # Paths to stratification files #
-  #################################
-
-  # file that relates the stratification to the Haul
-  # The two stratification types are found in two sheets: "Base KS" and "INPFC"
-  strata_filename: Stratification/US_CAN strata 2019_final.xlsx
-  strata_sheetname: Base KS
-
-  # file that defines the geographic definition of strata
-  # The two stratification types are found in two sheets: "stratification1" and "INPFC"
-  geo_strata_filename: Stratification/Stratification_geographic_Lat_2019_final.xlsx
-  geo_strata_sheetname: stratification1
-
-  # file that provides NASC values, which do not include age1 values
-  nasc_no_age1_filename: Exports/US_CAN_detailsa_2019_table2y+_ALL_final - updated.xlsx
-  nasc_no_age1_sheetname: Sheet1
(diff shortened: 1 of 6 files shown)


config_files/survey_year_2019_config.yml
```diff
@@ -4,94 +4,81 @@
...
+    filename: Exports/US_CAN_detailsa_2019_table1y+_ALL_final - updated.xlsx
+    sheetname: Sheet1
+kriging:
+  mesh:
+    filename: Kriging_files/Kriging_grid_files/krig_grid2_5nm_cut_centroids_2013.xlsx
+    sheetname: krigedgrid2_5nm_forChu
+  smoothed_contour:
+    # filename: Kriging_files/Kriging_grid_files/Smoothing_EasyKrig.xlsx
+    filename: Kriging_files/Kriging_grid_files/transformation_isobath_coordinates.xlsx
+    sheetname: Smoothing_EasyKrig
```
leewujung, line 79: Does this file contain the coordinate of the 250m isobath? Thinking if that's the case, maybe we can change the yaml item name from `smoothed_contour` to `isobath_250m` or something similar.

config_files/survey_year_2019_config.yml
```diff
@@ -4,94 +4,81 @@
...
-
-  # File names for the Canada section
-  filename_haul_to_transect_CAN: Biological/CAN/haul_to_transect_mapping_2019_CAN.xlsx
-  haul_to_transect_CAN_sheetname: Sheet1
-  ##############################################################################
+##############################################################################
+# Parameters
+
+survey_year: 2019            # survey year being considered
+species_id: 22500            # target species ID for the survey year
```
leewujung, line 11: What do you think about change this to
```ini
species: hake
```
and do the mapping under the hood? Maybe with a comment that the species name will be used to find a species ID under the hood?

config_files/survey_year_2019_config.yml
```diff
@@ -4,94 +4,81 @@
...
+CAN_haul_offset: 200         # The value to be added to the Canadian's haul number
+hemisphere: NW               # north/south and west/east hemispheres, used for the reports
```
leewujung, line 13: Do we need to retain this in the YAML? I guess I don't know how this is used.

leewujung (commented):
Hey @emiliom : Thanks for the PR. The revised YAML looks great! I just have a couple of small comments.

config_files/survey_year_2019_config.yml
emiliom: > Does this file contain the coordinate of the 250m isobath? 

Correct (see [here](https://uw-echospace.github.io/EchoPro/input-files.html#smoothed-shelf-break-contour)), except my understanding from the Hake Survey Methods document is that it's the 200m isobath, not 250m. 

Great suggestion! The name "smoothed contour" is unhelpful, but it's used widely, unfortunately. I agree with changing it here, in this user-facing context, and I like your suggested name -- So, `isobath_200m`?

Down the line, I think it'd be good to change variable names in the code to move away from "smooth contour". But let's do that in a separate PR.

config_files/survey_year_2019_config.yml
emiliom: You're referring to `hemisphere`, right? I had done a quick search through the code to see if it was used at all, and it isn't. I've confirmed this just now. So, I agree, let's drop it.

emiliom:
Thank you @leewujung ! I've commented on your inline comments.

In this PR I was focusing solely on the parameters that specify file paths and Excel sheet names. I'm happy to expand the scope to the stand-alone parameters at the top, that you commented on. 

One question for you: do you think we should group all the data-file parameters under a parameter like `hake_data`? Then, krill inputs would be under a header like `krill_data`. I can see pros and cons to doing that grouping.

config_files/survey_year_2019_config.yml
emiliom: I've pushed a commit removing `hemisphere`. Assuming you were referring only to that parameter, I'll flag this as resolved.

config_files/survey_year_2019_config.yml
emiliom: I've pushed an implementation of this where the mapping is in `survey_year_2019_config.yml`; we can easily move it to the other yaml if that's our consensus. 

I used the species name "pacific_hake" to be more specific.

config_files/survey_year_2019_config.yml
emiliom: I've pushed a commit implementing the change to `isobath_200m`

config_files/survey_year_2019_config.yml
leewujung: I like having this `species_id_mapping`, but I think it should be in `core.py` or something similar, since these are constant and should not be changed, if we do use the number 22500 under the hood.

In fact, do we have to use a number for the species? If we just use `species: pacific_hake` and use that in the code, wouldn't that be simpler?

config_files/survey_year_2019_config.yml
leewujung: Sounds good!

leewujung:
> One question for you: do you think we should group all the data-file parameters under a parameter like `hake_data`? Then, krill inputs would be under a header like `krill_data`. I can see pros and cons to doing that grouping.

I would prefer to not over-categorizing there, to make whatever that are in the yaml file not so hard-locked with the species. What I think we can do is to have yaml "template" for hake, and a different yaml "template" for krill. We can then add other species with data files that are relevant to those in the future. And of course with this approach we would have to have species-specific yaml checkers, for hake and krill, to start with.

config_files/survey_year_2019_config.yml
emiliom: We do need the code (22500 for hake) somewhere, at least at this time, because the biological data files currently include data for what I assume is all species in the hauls. `species_id`  is used to filter the data to retain only pacific hake.

I'd been thinking about this lately and I think we should the FEAT team about switching over to using biological input files that have data only for hake. That would simplify the code, too. But let's address that outside this PR.

I'll find an appropriate module for `species_id_mapping` and push a commit.

emiliom:
PR #126 (merged) created the conflicts now seen in this PR. I'll fix the PR after I resolve the conflicts locally.

leewujung:
Sounds good. I'll merge `main` to #123 once you fix this one, and we can merge that at that time.

I'll continue my reading still based on the current code without #126, and if anything identified there (#125) have already been resolved by #126 we can just tick the boxes. If you could merge this PR today that'll still help significantly with avoiding redundant effort. Thanks!

config_files/survey_year_2019_config.yml
emiliom: Done.

emiliom:
I've merged `main` and moved `species_id_mapping` dict from yaml file to `biological_data.LoadBioData`. I'll go ahead and merge this PR now. Thanks for your review!

PR #123, opened 2023-12-02
Use `setuptools_scm` for auto versioning

leewujung:
I ran into pip failure while using `condaenvironment.yaml` to create a mamba environment. The failure occurred either using the original `condaenvironment` with the pip entry, or doing `pip install -e .` separately. The messages suggested that it is likely not a problem of pip, so I did some searches. It seems that the combination in `setup.py` and `pyproject.toml` may be to blame, so I made some changes in this PR.

Note I removed the pip entry in `condaenvironment.yaml`, and do `pip install -e .` after the environment is created to install EchoPro in development mode. Including the pip entry still did not work and I did not look into why, because I think we should create a conda package for EchoPro. 

Also, the installation instructions are conflicting between installing as a regular user vs a developer. I did not attempt to modify it, since I believe the section should be revamped.


emiliom:
Thanks for creating this PR for moving to `setuptools_scm` and other package configuration improvements!

It looks like this PR also folds in what's in PR #121, though. Could you remove those extraneous changes? I think the changed files that should remain in this PR are `.flake8`, `.gitignore`, `EchoPro/__init__.py`, `condaenvironment.yaml`, `pyproject.toml` and `setup.cfg`. Thanks!!

leewujung:
Yeah, I merged that into this branch because I was starting to run through the code, and the first thing I ran into was not be able to create an environment. How about we merge #121 first and I'll pull those again? since the files changed here do not overlap with those.

emiliom:
> How about we merge #121 first and I'll pull those again? since the files changed here do not overlap with those.

Sounds good. I should be able to merge PR #121 shortly.

leewujung:
@emiliom: I just pulled in the latest merges in `main`, and this PR should be clean for review now.

emiliom (approved):
Thanks! Looks good. 

I see now that the pip entry in `condaenvironment.yaml` indeed should not have been there. Another mistake somewhat related to this is in the developer installation instructions in README.md. Instead of:
```bash
conda env create -f condaenvironment.yaml
```
It should be:
```bash
conda create -c conda-forge -n echopro python=3.9 --file requirements.txt
```

Should I push a commit into this PR to make that change?

leewujung:
> It should be:
> 
> ```shell
> conda create -c conda-forge -n echopro python=3.9 --file requirements.txt
> ```
> 
> Should I push a commit into this PR to make that change?

Sure. Thanks!

leewujung:
Actually, I need to submit another PR for a small data type change, so I will merge this now. Could you please do the installation update in another PR? Thanks!

Issue #95, opened 2023-09-26
Transect subset selection bug


emiliom:
PR #89 overhauled the interim transect subset selection approach initially implemented in [v0.1.0-alpha](https://github.com/uw-echospace/EchoPro/releases/tag/v0.1.0-alpha). The intent was to implement the approach found in Matlab EchoPro, as described in #33.

I have found a bug that occurs in some circumstances. The bug emerged when rerunning the [bootstrapping_walkthrough.ipynb](https://uw-echospace.github.io/EchoPro/example_notebooks/bootstrapping_walkthrough.html) notebook with the current `main`. PR #89 introduced a very small change to that notebook, changing the `removal_percentage` argument value in `boot_obj.run_bootstrapping` from 50.0 to 60.0. The notebook runs successfully when using 50.0, but `boot_obj.run_bootstrapping` results in a transect selection error when using 60.0

Currently the notebook [transect_selection_workflow.ipynb](https://uw-echospace.github.io/EchoPro/example_notebooks/transect_selection_workflow.html) runs successfully while also applying transect subsetting. But the percentage used is also 50.0%. In addition, there is a transect subsetting test, [test_transect_selection.py::test_transect_selection_output](https://github.com/uw-echospace/EchoPro/blob/a0d648af8267ac4b8d041e58361c25e4a24c9ec1/EchoPro/tests/computation/test_transect_selection.py#L8), that also runs successfully; rather than a %, it uses a preselected list of transect ids.

The error happens in `computation/transect_results.py::set_adult_NASC`, specifically in this statement:
https://github.com/uw-echospace/EchoPro/blob/9dc4708b409f9b0a71897dd06a690d26eb1a2e8d/EchoPro/computation/transect_results.py#L1272-L1274
[...]

PR #126, opened 2023-12-03
Several small but significant updates to input files and input file processing

emiliom:
- Rename `geo_strata` input `Latitude (upper limit)` column to `northlimit_latitude`, to sanitize the name as we've done with other columns. Updated the data file and the code.
- `transect_num` in haul_to_transect file is now handled as an `int` (rather than a float) and can not contain nulls. I edited the US file to remove rows where `transect_num` was null, and update the code. This leads to a more logical data type (int) and a small simplification of the code.
- Drop unused catch `haul_count` column.
- Set `sex` input column to `np.int8` data type instead of `int` to trim memory use.
- Replace `float` with `np.float64` for column data types in `kriging_mesh.py`, for consistency.
- Several updates to the input-files docs page, including populating 'Empty value' column.

Unrelated to input files per se, in this PR I also:
- Added skipping (`@pytest.mark.skip`) of `test_transect_selection.py::test_transects_selection_success` due to the partial failures documented in issue #95. Once #95 is addressed, we can remove this test skip.
- Fixed some Pandas warnings

### DIFF
--- a/EchoPro/tests/computation/test_transect_selection.py
+++ b/EchoPro/tests/computation/test_transect_selection.py
@@ -7,6 +7,8 @@ import EchoPro
 from EchoPro.computation import SemiVariogram as SV
 
 
+# 12/1/2023: The test typically succeeds for a few iterations, then fails
+@pytest.mark.skip(reason="Fails due to https://github.com/uw-echospace/EchoPro/issues/95")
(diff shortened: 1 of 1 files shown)


emiliom:
These are discrete changes that I'll go ahead and self merge.

Issue #113, opened 2023-10-08
Document core classes


emiliom:
### Each (major?) class

- Core class
  - survey.Survey
  - All actions and computations are orchestrated from this class
- computation
  - computation.transect_results.ComputeTransectVariables
  - computation.semivariogram.SemiVariogram
  - computation.kriging.Kriging
  - computation.kriging_variables.ComputeKrigingVariables
  - computation.bootstrapping.Bootstrapping
- data loaders
  - data_loader.kriging_mesh.KrigingMesh
  - data_loader.biological_data.LoadBioData
  - data_loader.stratification_data.LoadStrataData
- report generation
  - reports.Reports

### Other notes

- `bio_calc`: From `computation.transect_results.ComputeTransectVariables`; and also from Kriging?
  - Categorize the long list of df's and gdf's available on `bio_calc`; eg, `kriging`, `kriging_bin`, `transect`, `weight_fraction`, etc. These can be categorized either by prefix or suffix
  - Only the 6 "results" tables are gdf's: transect vs kriging, and sex (all[empty]/male/female). eg, `transect_results_male_gdf`
  - `bin_ds` is an xarray dataset
  - Most of the other objects are dataframes (`_df`), but there are a few others that don't have a clarifying suffix: `strata_sig_b`, `all_strata`, `missing_strata`, `percentage_transects_selected`, `sel_tran_strata_choice` (a dict), `stratum_choices` (a dict)
- For the classes, separate out public from private methods
- To get all the methods and attributes/properties of an object (a class), use this: `[(name, type(myobject.__getattribute__(name))) for name in dir(myobject) if not name.startswith("__")]`

emiliom:
I've been working on this. I'll finalize PR #131 on Friday.

leewujung:
I'll close this now as I just closed #131 and the data structure is being changed.

PR #131, opened 2023-12-04
docs: Expand API page and finalize Classes and data structures

emiliom:
- Expand docs API page to cover all classes and data loaders. The coverage was not complete.
- TODO: In later commits, will finalize the [Classes and data structures](https://uw-echospace.github.io/EchoPro/classes-datastructures.html) docs page (#113)

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -4,12 +4,19 @@ API reference
 **Content**
 
 * `Survey`_
-* `Computational routines`_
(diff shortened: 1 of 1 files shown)


leewujung:
I will go ahead to merge this now, since @brandynlucca has re-organized the data structure and will need to update the related doc content, and it will be nice to see how the previous API refs are setup and start revision with a clean state.

Issue #135, opened 2023-12-05
Why is type `"length_count": np.float64`?


leewujung:

emiliom:
>  Or is it because the entry could be empty?

That's correct.

Issue #133, opened 2023-12-05
Important expected input excel file changes


leewujung:
@emiliom : I am creating this issue because it took me a while to track down what happened, because I do not have the updated data files.

I think it is best to document this explicitly in an issue so that FEAT have a list for this. I started what I am running into here. Could you please add anything you remember below?

- the "haul_to_transect" file should NOT have rows that contain empty `transect_num` field
- the "Latitude (upper limit)" field in `geo_strata` input is renamed to "northlimit_latitude"

emiliom:
I introduced those two changes a few days ago, in PR #126.

> the "haul_to_transect" file should NOT have rows that contain empty `transect_num` field

The empty `transect_num` entries were a remnant of the that table originally coming from the gear file. By removing those rows, the data type specification became more logical (`transect_num` is an `int`) and the code a bit simpler.

> the "Latitude (upper limit)" field in `geo_strata` input is renamed to "northlimit_latitude"

We overlooked sanitizing this column name when we went through all others. Apparently Brandon overlooked the geo-strata file when we went through this, so we never discussed it. So, I sanitized it now to be consistent.

It's worth noting that we renamed several columns during that process. Old and new names, and most of our discussions, are in [this Google Sheet](https://docs.google.com/spreadsheets/d/18xMQOUd1dRquzzTOh3yi21-eOPqCnKsByaW2z1C4LI4/edit#gid=1860002679). I updated today the sheet listing new and old names, as it was missing some things and had a couple of outdated names. Should we add the old names as a new column in the tables in https://uw-echospace.github.io/EchoPro/input-files.html? With one or two exceptions, that page describes the current specifications, not how they have changed; but one more column doesn't seem like much clutter.

emiliom:
I can do that. But before moving forward, an alternative would be to place here, in this issue rather than in the inputs page, the tables of new vs old column names found in the Google Sheet.

What do you prefer?

leewujung:
Awesome, it looks very nice!

PR #147, opened 2023-12-06
Couple of updates to input-files doc page

emiliom:
- Add empty-value entry for length file `length_count` column. Follow up to #135
- Add more explicit clarification about haul-vs-transect `transect_num` column: "Note that rows with empty `transect_num` must be omitted." #133


emiliom:
Will self merge

PR #150, opened 2023-12-07
Input files page, add "Old name" column to description tables

emiliom:
Addresses #133


emiliom:
Will self merge, since we already discussed this in #133.

PR #152, opened 2023-12-10
Ensure new reports are actually generated when `test_reports.py` tests are run

emiliom:
In `test_reports.py`, the function `generate_reports` that generates the report Excel files was not actually set up to run automatically when a test was run; it had to be run manually in advance!
https://github.com/uw-echospace/EchoPro/blob/161163e293bd0024edd3fb1455a4103b3a213503/EchoPro/tests/reports/test_reports.py#L12

I've decorated the function with `@pytest.fixture(scope="module")` (see [here](https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-scopes)) and all the actual tests with `@pytest.mark.usefixtures("generate_reports")` (see [here](https://docs.pytest.org/en/latest/how-to/fixtures.html#use-fixtures-in-classes-and-modules-with-usefixtures)) to ensure `generate_reports` is called whenever a `test_reports.py` test is run. If multiple tests are run, `generate_reports` is run only once, with the first test; there is no need to run it again for every test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 _echopro_version.py
 
 # ignore output reports, if they were produced
-EchoPro/tests/reports/EchoPro_python_output
+EchoPro/tests/reports/EchoPro_python_output*
 
 # Created by .ignore support plugin (hsz.mobi)
 ### Python template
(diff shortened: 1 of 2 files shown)


emiliom:
I'll self merge. This PR contains an important fix to an obvious oversight.

Issue #153, opened 2023-12-10
Additional columns from NASC file are loaded in `reports.py` rather than in the main input data loading routines


emiliom:
It turns out the `reports.py` module re-loads the NASC file, this time bringing in 4 columns that were not loaded in the main data loading functions because they're not used in computations:

https://github.com/uw-echospace/EchoPro/blob/78ce64952ef1171a1fd3497cd0c58696bddf5db3/EchoPro/reports.py#L393-L400

Two of those columns (depth columns) are also used in some of the plots (see #145).

It would probably be best if these columns were loaded from the outset in the main data loading functions, so they're available for the report and plot generation. That would also eliminate the duplicated reading of the NASC file that's currently going on in `reports.py`. 

While addressing this, review the column names to "sanitize" them as previously done with all other input file columns (eg, replace blank spaces with underscores, make all lowercase whenever possible). Also decide if all four columns are actually needed. For the reports, I think they're being loaded only so the report files have an identical set of columns as the existing Matlab EchoPro report files.

leewujung:
@brandynlucca : tagging you here as it is relevant to the input data loader. See #154, which adds these into the documentation. I think for now we can load them in the data loader and produce the report files as is. Later once we are at a stage to verify all report files and figures, let’s work with FEAT to determine if some columns can to be trimmed.

brandynlucca:
This has been addressed by the changes incorporated into how `Echopop` ingests acoustic data via `Survey.load_acoustic_data()`. This can therefore be closed.

PR #154, opened 2023-12-10
In input files docs page, add NASC columns loaded in `reports.py`

emiliom:
Here's a first attempt at adding these "missing" 4 NASC-file columns (see #153) to the input-files docs page. @leewujung please change it as you see fit.


leewujung:
Thanks for adding this! Let’s merge this now so the content gets into the documentation.

Issue #128, opened 2023-12-04
Refactor functions called in `Survey.load_survey_data`


leewujung:
Below are functions and function calls that I think can be simplified.

I'll continue to add to this as I go, and will also create issues for additional functions to be added that are currently marked as TODOs.

- [x] #138
- [x] #136
- [x] #137
- [x] #132
- [x] #130
- [x] #149
- [x] #134
- [x] #135

brandynlucca:
> * [ ]  [Absorb `_process_nasc_data` into `load_nasc_df` #137](https://github.com/uw-echospace/EchoPro/issues/137)

I went ahead and checked some of these off since they've been addressed. For this one, it seems like this has been addressed (when perusing the linked Issues), but I wasn't sure if this is also inclusive of importing the separate NASC data files as a missing feature/functionality.

leewujung:
It wasn't intended to include importing the separate NASC data files, because I did not know about that at that time...

I just checked that box off also. I guess we're ready to close this! 🎉

I'll close those resolved issues that are still open also.

PR #160, opened 2024-01-02
Adds a `core.py` API at the root level

brandynlucca:
This addresses #128 whereby the `core.py` is defined at the root level of
the package and comprises two internal APIs: `CONFIG_MAP` and `LAYER_NAME_MAP`.
`CONFIG_MAP` is referenced for validating and enforcing both the required column names
and associated datatypes. `LAYER_NAME_MAP` is the current hard-coded
implementation that re-organizes the ``Survey`` class data attributes.

### DIFF
--- a/EchoPro/core.py
+++ b/EchoPro/core.py
@@ -1,0 +1,174 @@
+"""
+Core API/components for class structures, class methods, and utility/helper methods
+"""
+import numpy as np
+
+CONFIG_MAP = {
+    """
+    ``CONFIG_MAP`` defines the expected column names for all datasets defined within both
+    ``initialization_config.yml`` and ``survey_year_2019.yml``. This further pairs the
+    expected datatype with each variable. The nested dictionary names represent 
+    tags that are queried when the data are loaded which reflect the naming conventions
+    found within the configuration files.
+    """
+    # TODO: Update documentation to reflect dynamic parameterization since not all datasets
+    # will be relegated to just the year 2019 upon deployment.
+
+    # BIOLOGICAL DATASET -- LENGTH
+    "biological.length": {
+        "haul_num": int ,
+        "species_id": int ,
+        "sex": np.uint8 ,
+        "length": np.float64 ,
+        "length_count": int ,
+    } ,
+    # BIOLOGICAL DATASET -- CATCH
+        "biological.catch": {
+        "haul_num":  int ,
+        "species_id": int ,
+        "haul_weight": np.float64 ,
+    } ,
+    # BIOLOGICAL DATASET -- HAUL TO TRANSECT
+    "biological.haul_to_transect": {
+        "haul_num": int ,
+        "transect_num": int ,
+    } ,
+    # STRATIFICATION DATASET -- STRATA
+    "stratification.strata": {
+        "stratum_num": int ,
+        "haul_num": int ,
+        "fraction_hake": np.float64 ,
+    } ,
+    # STRATIFICATION DATASET -- GEOSTRATA
+    "stratification.geo_strata": {
+        "stratum_num": int ,
+        "northlimit_latitude": np.float64 ,
+    } ,
+    # ACOUSTIC DATASET -- NASC EXCLUDING AGE-1
+    "NASC.no_age1": {
+        "transect_num": int ,
+        "vessel_log_start": np.float64 ,
+        "vessel_log_end": np.float64 ,
+        "latitude": np.float64 ,
+        "longitude": np.float64 ,
+        "stratum_num": int ,
+        "transect_spacing": np.float64 ,
+        "NASC": np.float64 ,
+        "haul_num": int ,       
+    } ,
+    # ACOUSTIC DATASET -- NASC ALL AGES
+    "NASC.all_ages": {
+        "transect_num": int ,
+        "vessel_log_start": np.float64 ,
+        "vessel_log_end": np.float64 ,
+        "latitude": np.float64 ,
+        "longitude": np.float64 ,
+        "stratum_num": int ,
+        "transect_spacing": np.float64 ,
+        "NASC": np.float64 ,
+        "haul_num": int ,       
+    } ,
+    # STATISTICAL DATASET -- KRIGING MESH
+    "kriging.mesh": {
+        "centroid_latitude": np.float64 ,
+        "centroid_longitude": np.float64 , 
+        "fraction_cell_in_polygon": np.float64 ,    
+    } ,
+    # STATISTICAL DATASET -- 200 M ISOBATH
+    "kriging.isobath_200m": {
+        "centroid_latitude": np.float64 ,
+        "centroid_longitude": np.float64 , 
+        "fraction_cell_in_polygon": np.float64 ,  
+    } ,
+    # STATISTICAL DATASET -- KRIGING VARIOGRAM PARAMETERIZATION
+    "kriging.vario_krig_para": {
+        "dataprep.y_offset": np.float64 ,
+        "vario.range": np.float64 ,
+        "vario.res": np.float64 ,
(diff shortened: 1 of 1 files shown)


EchoPro/core.py
```diff
@@ -0,0 +1,174 @@
...
+    ``initialization_config.yml`` and ``survey_year_2019.yml``. This further pairs the
+    expected datatype with each variable. The nested dictionary names represent 
+    tags that are queried when the data are loaded which reflect the naming conventions
+    found within the configuration files.
+    """
+    # TODO: Update documentation to reflect dynamic parameterization since not all datasets
+    # will be relegated to just the year 2019 upon deployment.
+
+    # BIOLOGICAL DATASET -- LENGTH
+    "biological.length": {
```
leewujung, line 18: The use of `.` is pretty common in R, but I think it would be good to stick with "_" for variables in Python.
```suggestion
    "biological_length": {
```

EchoPro/core.py
```diff
@@ -0,0 +1,174 @@
...
+    Each dictionary comprises the same dataset names contained within the configuration file.
+    Nested within each of these dictionaries include three variables: 
+        name (string): the name of the attribute appended to a ``Survey`` class object.
+        data (list[string]): the name of the expected data layer, which is formatted as a dataframe.
+        superlayer (list[string]): layer names that are prepended onto the nested dictionary structure of the data attribute.
+    While this reference library is only required in the sense that it is hard-coded into the
+    the current implementation concerning loading data into a ``Survey`` class object, it may be 
+    helpful for organizing data with more parsimonious and representative names. It may also be
+    helpful for future extensions enabling users to set additional data attributes that aren't 
+    included in the ``Survey`` class initialization and parameterization (when the object is generated).
```
leewujung, line 150: Just trying to get it into a format that is similar to numpydoc, though these won't really be rendered.
```suggestion
    ``LAYER_NAME_MAP`` is a hard-coded dictionary that aids in re-mapping the data tree
    structure of various attributes contained within a generated ``Survey`` class object. 
    Each dictionary comprises the same dataset names contained within the configuration file.
    
    Nested within each of these dictionaries include three variables: 
        name : string)
            The name of the attribute appended to a ``Survey`` class object.
        data : list[string]
            The name of the expected data layer, which is formatted as a dataframe.
        superlayer : list[string]
            Layer names that are prepended onto the nested dictionary structure of the data attribute.

    While this reference library is only required in the sense that it is hard-coded into the
    the current implementation concerning loading data into a ``Survey`` class object, it may be 
    helpful for organizing data with more parsimonious and representative names. It may also be
    helpful for future extensions enabling users to set additional data attributes that aren't 
    included in the ``Survey`` class initialization and parameterization (when the object is generated).
```

EchoPro/core.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    # TODO: This is a hard-coded feature and therefore is not particularly helpful for more dynamic
+    # use of this Python module (and the overall use of the EchoPro package).
+    'biological': {
```
leewujung, line 154: Once you set up pre-commit, which uses the `.pre-commit-config.yaml` that's already in the repo, you'll find that `black` would automatically change all `'` to `"` -- I remember feeling all the double quotes to be crowding the view, but I guess that's black's choice :P and I've gotten used to it, ha.
```suggestion
    "biological": {
```

leewujung (commented):
Hey @brandynlucca : Thanks for adding this! I added some inline comments that are pretty small. Once these are changed we can merge this.

Also, since you're PRing to `main`, do you wanna change the branch name to something like `add_core`?

EchoPro/core.py
brandynlucca: Perhaps this may make more sense in the context of how I implemented `CONFIG_MAP` for `Survey` parameterization, but these dictionary key names with `CONFIG_MAP` are primarily used as tags to help link `CONFIG_MAP` to a specific data file for validation (e.g. length data for US fish). So in this case, text on each side of the `.` delimitor correspond to first two nested layers as defined in `survey_year_2019_config.yml`. So for instance: `'biological.length.US'` maps to the configuration settings that yield: `{'filename': 'Biological/US/2019_biodata_length.xlsx', 'sheetname': 'biodata_length'}`. Similarly, `biological.length` links the values within `CONFIG_MAP` to all files associated with `biological.length`, so in this case: `biological.length.US` and `biological.length.CAN`. I elected to use `.` over `_` since some nested dictionary key names also contain `_`, such as `all_ages'`. So `NASC.all_ages`, as an example, was easier to parse/separate than `NASC_all_ages` (at least with how the code is currently written) when needing to toggle between `NASC.all_ages` and `[NASC, all_ages]`. Since these tags are really representing the dictionary paths of the stored dataframes, perhaps it would make more sense to replace `.` with `/`? For reference, I have included a rough schematic of the data attribute structure:

* 'acoustics': dict
    * 'nasc': dict
        * 'no_age1_df': DataFrame, 
        * 'all_ages_df': DataFrame,
* 'biology': dict 
    * 'catch_df': DataFrame
        * ['CAN', 'US']: regions within df
    * 'haul_to_transect_df': DataFrame
        * ['CAN', 'US']: regions within df
    * 'length_df': DataFrame 
        * ['CAN', 'US']: regions within df
    * 'specimen_df': DataFrame 
        * ['CAN', 'US']: regions within df
* 'spatial': dict
    * 'strata_df': DataFrame, 
    * 'geo_strata_df': DataFrame,
* 'statistics': dict
    * 'kriging': dict 
        - 'mesh_df': DataFrame,
        - 'isobath_200m_df': DataFrame,
        - 'vario_krig_para_df': DataFrame

So if I wanted to get the NASC exports for all age-classes, the nested dictionary path would be represented by either: `'acoustics/nasc/all_ages_df'`,`'acoustics.nasc.all_ages_df'`, or `'acoustics_nasc_all_ages_df'`. Similarly, `'biology.length_df'` in the above structure would get the concatenated DataFrame containing both the `US` and `CAN` datasets. I have no problem with using `_` instead; however, it would require some tweaking to either the `survey_year_2019_config.yml` or the code that define, initialize, and parameterize `Survey` in `survey.py`. With all of this in mind, I personally think `/` may be a better delimiter in this case since it is a bit more intuitive to associate `/` with file paths (or in this case the dictionary paths) versus `.` or `_`.

EchoPro/core.py
brandynlucca: I did not realize that -- that is nifty! I went ahead and replaced `"` with `'` throughout, so I will mark this as resolved.

EchoPro/core.py
leewujung: I see, I can see the problem with having everything with `_`. 

If your goal is to map to what's in `survey_year_2019_config.yml`, how about using the same nested dictionary structures as what would be read in using `yaml.load` (currently used under `_read_config)`:
```python
CONFIG_MAP = {
    "biological": {
        # BIOLOGICAL DATASET -- LENGTH
        "length" : {
            "haul_num": int ,
            "species_id": int ,
            "sex": np.uint8 ,
            "length": np.float64 ,
            "length_count": int ,
        },
        # BIOLOGICAL DATASET -- CATCH
        "catch": {
            "haul_num":  int ,
            "species_id": int ,
            "haul_weight": np.float64 ,
        } ,
    },
    "stratification": {
        "strate" : {
            ...
        },
        "geo_strata": {
            ...
        }
    },
    ...
}
```
So that you could use exactly the same string to index both.

One of the recent changes Emilio did (in https://github.com/uw-echospace/EchoPro/pull/121) was exactly to enable splitting strings with `/` so that the `survey_year_2019_config.yaml` inputs is more succinct. So it could probably be that the underlying structure in `core` is the nested dictionaries, but when you write it in code, it would just be the strings with `/` delimiters with the same kind of splitting when it is used.

PR #161, opened 2024-01-05
Update core.py and begin refactoring data loader

brandynlucca:
Updated `core.py` to contain the expected nested dictionary data structures of
the imported data attributes.

The initialization of `Survey` includes importing
the configuration files (`initialization_config.yml` and `survey_year_2019_config.yml`)
and all associated data. This includes a utility function `populate_tree` that
maps out the current data attribute dictionary paths (e.g. where all of the
biological data is stored within the `Survey` class object). This is called by
the user via `Survey.summary`.

Minor adjustments were made to the configuration files.

Some support scripts have been added within `data_structure_utils.py` that
aid in pushing/pulling files from nested dictionaries.

### DIFF
--- a/EchoPro/survey.py
+++ b/EchoPro/survey.py
@@ -1,17 +1,15 @@
-from pathlib import Path
 from typing import List, Union
 from pathlib import Path
 import pandas as pd
 import numpy as np
 import yaml
 import os
-from openpyxl import load_workbook
-from .computation.acoustics import ts_length_regression, to_linear
 from .core import CONFIG_MAP, LAYER_NAME_MAP
 ### !!! TODO : This is a temporary import call -- this will need to be changed to 
 # the correct relative structure (i.e. '.core' instead of 'EchoPro.core' at a future testing step)
 import pprint
-from .utils.data_structure_utils import pull_nested_dict, push_nested_dict
+from .utils.data_structure_utils import push_nested_dict
+from .utils.data_file_validation import validate_data_columns
 ### !!! TODO : This is a temporary import call -- this will need to be changed to 
 # the correct relative structure (i.e. '.utils.data_structure_utils' instead of 
 # 'EchoPro.utils.data_structure_utils' at a future testing step)
@@ -54,7 +52,7 @@ class Survey:
         ### Loading the configuration settings and definitions that are used to 
         # initialize the Survey class object
         # ATTRIBUTE ADDITIONS: `config`
-        self.config = self.load_configuration(init_config_path, survey_year_config_path)
+        self.config = self.load_configuration( init_config_path , survey_year_config_path )
 
         ### Loading the datasets defined in the configuration files
         # EXAMPLE ATTRIBUTE ADDITIONS: `biology`, `spatial`, `acoustics`
@@ -63,20 +61,14 @@ class Survey:
         # Define length and age distributions
         self.biometric_distributions()
 
+        ### !!! THIS IS TEMPORARY FOR DEBUGGING / TRACKING DATA ATTRIBUTE ASSIGNMENT 
         ### A utility function that helps to map the datasets currently present
         # within Survey object via the `___.summary` property. This also initializes
         # the `meta` attribute
         # ATTRIBUTE ADDITIONS: `meta`
         ##
-        # TODO: Fix bug associated with `load_survey_data()` that results in the 
-        # 'dict_tree' key within the `___.config` attribute overwriting previous
-        # entries when multiple datasets of the same layer name are present, such as
-        # 'US' versus 'CAN' for biological data. At present, it defaults to listing 
-        # only 'CAN' -- this is not reflected in the actual data attribute structure where
-        # both 'US' and 'CAN' are present, denoted by the 'index' column in relevant dataframes.
         self.populate_tree()
 
-        
     @staticmethod
     def load_configuration( init_config_path: Path , 
                             survey_year_config_path: Path ):
@@ -87,6 +79,7 @@ class Survey:
 
         Parameters
         ----------
+        self : EchoPro.Survey
         init_config_path : pathlib.Path
             A string specifying the path to the initialization YAML file
         survey_year_config_path : pathlib.Path
@@ -102,100 +95,36 @@ class Survey:
         # Retreive the module directory to begin mapping the configuration file location
         current_directory = os.path.dirname(os.path.abspath(__file__))
 
-        # Build the full configuration file paths
+        # Build the full configuration file paths and verify they exist
         config_files = [init_config_path, survey_year_config_path]
-        config_file_paths = [os.path.join(current_directory, '..', 'config_files', file_name) for file_name in config_files]
-
-        # Check files to ensure they exist
-        config_existence = [os.path.exists(file_path) for file_path in config_file_paths]
-
-        # Assign the existence status to each configuration file for error evaluation
-        config_status = dict(zip(config_files, config_existence))
+        config_paths = [ Path(current_directory + '/../config_files/' + file) for file in config_files ]
+        config_existence = [path.exists() for path in config_paths]
 
         # Error evaluation and print message (if applicable)
-        if not all(config_status.values()):
-            missing_config = [config_file for config_file, exists in config_status.items() if not exists]
+        if not all(config_existence):
+            missing_config = [ files for files, exists in zip( config_files, config_existence ) if not exists ]
             raise FileNotFoundError(f"The following configuration files do not exist: {missing_config}")
 
         ### Read configuration files
         # If configuration file existence is confirmed, proceed to reading in the actual files
         ## !!! TODO: Incorporate a configuration file validator that enforces required variables and formatting
-        init_config_params = yaml.safe_load(Path(config_file_paths[0]).read_text())
-        survey_year_config_params = yaml.safe_load(Path(config_file_paths[1]).read_text())        
+        init_config_params = yaml.safe_load(config_paths[0].read_text())
+        survey_year_config_params = yaml.safe_load(config_paths[1].read_text())        
 
         # Validate that initialization and survey year configuration parameters do not intersect
-        param_intersect = set(init_config_params.keys()).intersection(set(survey_year_config_params.keys()))
+        config_intersect = set(init_config_params.keys()).intersection(set(survey_year_config_params.keys()))
 
         # Error evaluation, if applicable
-        if param_intersect:
+        if config_intersect:
             raise RuntimeError(
-                f"The initialization and survey year configuration files comprise the following intersecting variables: {param_intersect}"
+                f"The initialization and survey year configuration files comprise the following intersecting variables: {config_intersect}"
             )
 
         ### Format dictionary that will parameterize the `config` class attribute
         # Join the initialization and survey year parameters into a single dictionary
-        full_params = {**init_config_params, **survey_year_config_params}
+        # Pass 'full_params' to the class instance
+        return {**init_config_params, **survey_year_config_params}
 
-        # ??? TODO: Since this is a static method, this carriage return may not be the most appropriate way to pass 
-        # 'full_params' to the class instance
-        return full_params
-    
-    def search_nested_dict( self ,
-                            config: dict ,
-                            filename: str ,
-                            current_path: list = [] ):
-        """
-        Recursive utility function that searches the configuration attribute
-        (a nested dictionary) for a specific 
-
-        Parameters
-        ----------
-        self
-            Survey class object.
-        config : dict
-            Survey class configuration ('self.config') dictionary.
-        filename : str
-            Target filename that is being searched within the configuration dictionary.
-        current_path : list
-            List that records previously evaluated levels/layers of the configuration
-        dictionary and steps forward to the next nested level within the dictionary.
-
-        Returns
-        -----
-        data_layer_names : list
-            List comprising the names of nested dictionary layers.
-        reference_path : str
-            A string representing the configuration tag that is referenced in the 
-        CONFIG_MAP API. 
-        sheet_name : str
-            References the correct sheet name containing the data that is defined
-        in the configuration files.
-        """
-        ### Initialize 'current_path' if not already defined/initialized
-        if current_path is None:
-                current_path = []
-
-        ### Recursive function 
-        # Step forward through the configuration dictionary until the filename input
-        # is located.
-        for key, value in config.items():
-            # Not located ? Keep searching
-            if isinstance(value, dict):
-                result = self.search_nested_dict(value, filename, current_path + [key])
-                if result is not None:
-                    return result
-            # Located ? Provide the outputs
-            elif key == 'filename' and value in filename:
-                return current_path,'/'.join(current_path[0:2]),config.get('sheetname')
-            
-        ### Never present !? 
-        ## !!! TODO : In theory, the 'load_configuration' function should validate that 
-        # the information required for the output is present. However, if that leaks through
-        # any cracks, then this becomes an unchecked recursive function. Consequently, this
-        # should be leashed so that it can only dig for so long until it yields some sort of 
-        # Value or Runtime Error. 
-        return None     
-    
     def load_survey_data( self ):
         """
         Loads the biological, NASC, and stratification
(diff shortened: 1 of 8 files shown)


config_files/initialization_config.yml
```diff
@@ -19,14 +25,24 @@
...
-                                        # will give coef = 10^(sig_b_coeff_power)
   #####################################################################################################################
 
+  #####################################################################################################################
+  # Log-linear regression#
+  ########################
+  # Target strenth (TS) - length (L) regression: TS=m*log10(L)+b
+  TS_length_regression_parameters:
+    pacific_hake:                         # corresponding species text code
+      number_code: 22500                  # species number code
```
leewujung, line 36: I wonder if it'll be good to have another place where the species name and code mapping is provided. It seems should be separate from the TS-length relationship and the config yamls.

leewujung: Actually now thinking about it, maybe there could be a place where the various type of codes, like the code for different sex could also be put into that other place...

config_files/survey_year_2019_config.yml
```diff
@@ -8,13 +8,22 @@
...
+    number_code: 1
+  Female:
+    text_code: F
+    number_code: 2
 
 ##############################################################################
 # Directory path that contains all input data needed
 
-data_root_dir: ../../2019_consolidated_files/
+data_root_dir: C:/Users/Brandyn/Documents/GitHub/EchoPro_data/2019_consolidated_files/
```
leewujung, line 26: You could change this to symbolic link, so that it always lives in the same place in the repo but could be linked to other places. In my local repo I have a symlink `EchoPro/data` that goes to some place where I put the test data.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+        ### Format dictionary that will parameterize the `config` class attribute
+        # Join the initialization and survey year parameters into a single dictionary
+        full_params = {**init_config_params, **survey_year_config_params}
 
-        return semi_vario
-
-    def get_kriging(self, params: krig_param_type) -> Kriging:
+        # ??? TODO: Since this is a static method, this carriage return may not be the most appropriate way to pass 
+        # 'full_params' to the class instance
+        return full_params
```
leewujung, line 141: I think this is fine?

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
 
-        Parameters
-        ----------
-        krig_mesh : KrigingMesh
-            Object representing the Kriging mesh
-        params : dict
-            Semi-variogram specific parameters. Contains the following
-            parameters:
+        # Check files to ensure they exist
+        config_existence = [os.path.exists(file_path) for file_path in config_file_paths]
```
leewujung, line 110: Since both `init_config_path` and `survey_year_config_path` are already `Path` objects, maybe you could just use the functionalities of `Path` in place for the `os.path` ops.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+    
+    def search_nested_dict( self ,
+                            config: dict ,
+                            filename: str ,
+                            current_path: list = [] ):
         """
-        Initializes a ``Kriging`` object using the
-        provided parameters
+        Recursive utility function that searches the configuration attribute
+        (a nested dictionary) for a specific
```
leewujung, line 149: Since the config dictionaries are clearly defined and under our control, the use of recursive function seems a bit overkill -- I think writing functions to search for specific files would suffice, especially since this function is used specifically for finding where a path is located, which is very specific already.

EchoPro/survey.py
leewujung, line 110: Since both `init_config_path` and `survey_year_config_path` are already Path objects, you can use the functionalities of `Path` in place for the `os.path` ops.

Also, I think this loop `data_file_paths = [os.path.join(self.config['data_root_dir'], filename) for filename in parsed_filenames]` can be omitted by checking `(Path(self.config['data_root_dir']) / filename).exists()` in the loop that produces `data_existence`.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+
+            # Determine how deep 'data_layers' is represented within the dictionary
+            # This is required for mapping dataframes with an appropriately named
+            # data layer name appended with '_df' (e.g. 'length' -> 'length_df')
+            data_name_bool = set(data_layers).intersection(layer_config['data'])
+            data_layer_index = data_layers.index(''.join(data_name_bool))
+
+            #pull_nested_dict(CONFIG_MAP, data_layers)
+            # Retrieve/access CONFIG_MAP values defining the expected column names and data types
+            validation_settings = pull_nested_dict(CONFIG_MAP, data_layers[:data_layer_index+1])
```
leewujung, line 251: The two mapping dict `LAYER_NAME_MAP` and `CONFIG_MAP` both have simple and known structures, and I think it is overkill to have dedicated functions like `search_nested_dict` and `pull_nested_dict` to find where some attributes are stored. These are essentially re-doing what XML tree was trying to achieve, and there are libraries that handles that already (note I am _not_ suggesting that you use XML! that would also be overkill).

I suggest that you just hard-coded the part about looking for `filename` and `sheetname`, given the known structures in the mappings. The code will be much easier to follow and debug/maintain later, since it makes use of the hierarchical structure in the yaml file.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+                workbook.close()
+
+                # Error evaluation and print message (if applicable)
+                if not set(validation_settings.keys()).issubset(set(data_columns)):
+                    missing_columns = set(validation_settings.keys()) - set(data_columns)
+                    raise ValueError(f"Missing columns in the Excel file: {missing_columns}")
+                
+            except Exception as e:
+                print(f"Error reading file '{file}': {e}")
+                continue
```
leewujung, line 279: I suggest factoring this part out to a function that checks for column name, to keep the main code flow in `load_survey_data` clean and easy to follow.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+                df = df.apply(lambda col: col.astype(validation_settings.get(col.name, type(col[0]))))
+
+                # This is a goofy way of accounting for the 'CAN' and 'US' format -- if these are present,
+                # they are added as an 'index' column that can be referenced later on.
+                ## ??? TODO : This janky boolean chunk can be replaced with something more efficient
+                # such as adding a key to the original configuration files that explicitly state datasets
+                # from multiple sources/regions/etc. are present, or automatically merge layers with shared 
+                # data tree structures in a later function/step
+                if len(data_layers) > 2:
+                    df['index'] = data_layers[2]
```
leewujung, line 319: I suggest factoring this part to a function that returns the organized `df`.

EchoPro/survey.py
```diff
@@ -36,741 +25,457 @@ class Survey:
...
+        data_status = dict(zip(data_file_paths, data_existence))
+
+        # Error evaluation and print message (if applicable)
+        if not all(data_status.values()):
+            missing_data = [data_file for data_file, exists in data_status.items() if not exists]
+            raise FileNotFoundError(f"The following data files do not exist: {missing_data}")
+        
+        ### Data validation and import
+        # Iterate through each file defined/detected in 'data_file_paths'
+        for file in data_file_paths:
```
leewujung, line 231: I read through what's in this entire for loop, and think that the code would be a lot cleaner if you do not try to write everything in a "one size fits all way" to check for layer depth and whether there's another layer above or below. Instead, I suggest hard coding which dict will be looking for `filename` and `sheetname` at the 3rd level (`biological`) and which dict will be looking for these at the 2nd level (`stratification`, `NASC`, `kriging`). This way the handling of that extra US/CAN level for the `biological` dict will also be explicit (and can be documented in the code comment and docstring). 

The two mapping dict `LAYER_NAME_MAP` and `CONFIG_MAP` both have simple structures and are completely under our control, and I think it is overkill to have the dedicated search/push/pull functions to deal with the 2nd or 3rd level indexing. These functions are doing something libraries that deals with XML trees are trying to achieve, but the "tree" structure we have here is small, I suggest to hard-code things when needed.

Along with factoring out 2 functions below ([here](https://github.com/uw-echospace/EchoPro/pull/161/files#r1443609137) and [here](https://github.com/uw-echospace/EchoPro/pull/161/files#r1443609380)), with a small amount of hard-coded selection rules, I think the code will be much easier to follow and debug/maintain later, since it directly makes use of the hierarchical structure in the yaml file.

leewujung (commented):
Hey @brandynlucca : Thanks for creating the draft PR, it's great to see the code!

As you will see in my inline comments, my main suggestions boil down to: hard-coding things is sometimes a good thing, when the "tree" structure we are dealing with is simple and completely under our control -- so make use of it to make the code easier to follow (and therefore easier to maintain/debug later).

Related to this, in a couple places I saw that the hierarchical structure from the yaml was flatten out so allow simple filtering. I didn't make comments because I felt the first use in `load_survey_data` does make it easy to get the filenames, and I did not read through the case in `populate_tree` since it is part of the tree-handling stuff that I commented on above.

I also made a couple comments about things that can be factored out to be separate checking/loading functions, which should be pretty straightforward.

EchoPro/survey.py
leewujung: See my other more overarching comment https://github.com/uw-echospace/EchoPro/pull/161#discussion_r1443614042

config_files/initialization_config.yml
leewujung: I think these can be in `core.py` along with other definitions:
```python
SEX_CODE_TEXT = {
   "male": "M",
   "female": "F",
   "unsexed": "U",  # or just an empty string "", depending on what's in the spreadsheet
}

SEX_CODE_TEXT = {
   "male": 1,
   "female": 2,
   "unsexed": 3, 
}

SPECIES_CODE = {
   "hake": 22500,
   "krill": ...,
}
```

brandynlucca:
@leewujung 
I pushed a good chunk of changes we discussed [here](https://github.com/uw-echospace/EchoPro/pull/161/commits/579060bf6e99cf8eba76859a445af5dbf65ccb1a). I've also copied-pasted the commit message below for reference. Similar as before, this succeeds in all steps between initializing the `EchoPro.Survey` object through populating and formatting all of the data attributes. 

- Reduced the number of recursive functions
-- Maintained the `populate_tree()` function solely for debugging purposes. Its functionality is completely independent of the rest of the module in its current state.
--- This concerns: `PARASITE_TREE`, `pushed_nested_dict`, `add_trees`, `populate_tree`

- Factored out the column validation
-- Now located at `EchoPro.utils.data_file_validation`
--- This concerns: `validate_data_columns`

- Factored out data type validation and import
--- This concerns: `read_validated_data`
--- This can probably be further consolidated with a little more hard-coding given that it started from a recursive-heavy state.

EchoPro/survey.py
```diff
@@ -87,6 +79,7 @@ def load_configuration( init_config_path: Path ,
 
         Parameters
         ----------
+        self : EchoPro.Survey
```
leewujung, line 82: Usually one can omit `self` in docstrings.

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
-
-            - 'all' -> loads all survey data
-            - 'biological' -> only loads the biological data
-            - 'strata' -> only loads the stratification data
-            - 'nasc' -> only loads the NASC data
+        self : EchoPro.Survey
+        init_config_path : pathlib.Path
+            A string specifying the path to the initialization YAML file
+        survey_year_config_path : pathlib.Path
+            A string specifying the path to the survey year YAML file
```
leewujung, line 86:
```suggestion
        init_config_path : pathlib.Path
            A Path object specifying the path to the initialization YAML file
        survey_year_config_path : pathlib.Path
            A Path object specifying the path to the survey year YAML file
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+        becomes available for future reference and functions.
+        """
+        ### Validate configuration files
+        # Retreive the module directory to begin mapping the configuration file location
+        current_directory = os.path.dirname(os.path.abspath(__file__))
+
+        # Build the full configuration file paths and verify they exist
+        config_files = [init_config_path, survey_year_config_path]
+        config_paths = [ Path(current_directory + '/../config_files/' + file) for file in config_files ]
+        config_existence = [path.exists() for path in config_paths]
```
leewujung, line 101: The input paths should be completely independent from the module directory, since they should be either relative path to the current directory of wherever the user is, or absolute path. `Path` handles both.

Below is my suggestion:
```suggestion
        # Build the full configuration file paths and verify they exist
        config_existence = [init_config_path.exists(), survey_year_config_path.exists()]
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+
+        ### Check whether data files defined from the configuration file exists
+        # Generate flat JSON table comprising all configuration parameter names
+        flat_configuration_table = pd.json_normalize(self.config).filter(regex="filename")
+
+        # Parse the flattened configuration table to identify data file names and paths
+        parsed_filenames = flat_configuration_table.values.flatten()
+
+        # Append the root directory to the parsed data file names
+        data_paths = [Path(self.config['data_root_dir'] + '/' + file) for file in parsed_filenames]
```
leewujung, line 145: `Path` objects can handle string directly:
```suggestion
        data_paths = [Path(self.config['data_root_dir']) / file for file in parsed_filenames]
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+
+        # Evaluate whether either file is missing
+        data_existence = [path.exists() for path in data_paths]
```
leewujung, line 148: `Path` objects can handle concatenation directly, and the list comprehension can be condensed into one.
```suggestion
        # Evaluate whether either file is missing
        data_existence = [(Path(self.config['data_root_dir']) / file).exists() for file in parsed_filenames]
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+
+        # Assign the existence status to each configuration file for error evaluation
+        # Error evaluation and print message (if applicable)
+        if not all(data_existence):
+            missing_data = [ files for files, exists in zip( parsed_filenames, data_existence ) if not exists ]
```
leewujung, line 153: You can index the list directly:
```suggestion
            missing_data = parsed_filenames[data_existence]
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+            raise FileNotFoundError(f"The following data files do not exist: {missing_data}")
+        
+        # Initialize data attributes ! 
+        setattr( self , 'acoustics' , { } )
+        setattr( self , 'biology' , { } )
+        setattr( self , 'spatial' , { } )
+        setattr( self , 'statistics' , { } )
```
leewujung, line 160: These should happen in `__init__`, usually in the form of:
```
self.acoustic = {}
self.biology = {}
self.spatial = {}
self.statistics = {}
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+        ### Data validation and import
+        # Iterate through each file defined/detected in 'data_paths' and 'parsed_filenames'
+
+        for file , path in zip(parsed_filenames, data_paths):
+
+            # Parse config to find appropriate settings
+            config_name = flat_configuration_table.apply(lambda row: row[row == file].index, axis=1)
+            
+            # Convert to string
+            config_identifier = config_name[0].tolist()[0].lower()
```
leewujung, line 171: I don't think you need to use apply for simple row selection? I have not tried to evaluate this, because there's path issue when I try to run the code, but I suspect if you use simple selection you would not have to do hard code `[0]` type of selection.

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+
+            # Extract appropriate dataset name from the file path
+            # -- Based on tail-end of file path string to avoid potential 
+            # -- directory name issues
+            dataset = [x for x in [*CONFIG_MAP.keys()] if x.lower() in config_identifier][0]
+            
+            # Determine data layer name
+            datalayer = [x for x in [*CONFIG_MAP.get(dataset)] if x in config_identifier.split('.')][0]
```
leewujung, line 179: You can change the loop to be 
```python
for data_type in ["biological", "stratification", "NASC", "kriging"]:
   # proceed to set data_layer
   # -- simpler since you know what it SHOULD be depending on data_type
```
This way the code in L167-179 can be mostly removed.

leewujung: And then I would loop through the regions
```
for region_id in ["US", "CAN"]:
   # Build config_map as in L186-187.
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
+
+            # Define the data layer name 
+            # -- Based on the lattermost portion of the file path string
+            # Create list for parsing the hard-coded API dictionary
+            if dataset == 'biological':                
+                region_id = [x for x in ['US', 'CAN'] if x.lower() in config_identifier][0]
+                sheet_name = self.config.get(dataset).get(datalayer).get(region_id).get('sheetname')
```
leewujung, line 186: This makes the code more readable because it is easy to directly see the multi-layer dict structure:
```suggestion
                sheet_name = self.config[dataset][datalayer][region_id]['sheetname']
```

EchoPro/survey.py
```diff
@@ -36,741 +23,440 @@ class Survey:
...
-        return krig
-
-    def compute_kriging_results(self, kriging_params: dict) -> None:
+            attribute_name = LAYER_NAME_MAP.get(config_map[0])['superlayer'][0]
+            layer_name = LAYER_NAME_MAP.get(config_map[0])['name']
+            internal = getattr( self , attribute_name )
+
+            # Add original data layer (e.g. stratification, kriging, etc.) if needed
+            if config_map[0] not in internal.keys():
+                internal = internal.setdefault(layer_name, {})
```
leewujung, line 295: This condition should never happen, since these should be initialized already in `__init__` (see https://github.com/uw-echospace/EchoPro/pull/161/files?diff=unified&w=0#r1449843433).

EchoPro/survey.py
leewujung: I see that `data_paths` are used later, so not condensing is fine.

EchoPro/survey.py
brandynlucca: Ah I see. Previously, I was just using the filename itself as the input (e.g. `initialization_config.yml`) since I assumed that this would not change in terms of the directory location within `EchoPro`. I am fine with maintaining the same input structures as shown in the `echopro_workflow.ipynb` and `transect_selection_workflow.ipynb` example notebooks.

brandynlucca: This appears to now work with the absolute path (e.g. `'C:/Users/Brandyn/Documents/GitHub/EchoPro/config_files/initialization_config.yml'`) and relative path (e.g. `./config_files/initialization_config.yml`), so I will resolve this conversation.

EchoPro/survey.py
brandynlucca: For reference, I am trying to retrieve the column name that matches the parsed_filename (e.g. `'Biological/US/2019_biodata_length.xlsx'`). So when running at least these two lines, they yield:
`config_name = flat_configuration_table.apply(lambda row: row[row == file].index, axis=1)`
---> `Index(['biological.length.US.filename'], dtype: object`

`config_identifier = config_name[0].tolist()[0].lower()`
---> `'biological.length.us.filename'`

When I use simple selection, this is certainly possible such that I select the columns directly:
`flat_configuration_table.columns[flat_configuration_table.isin([file]).any()]` ---> `Index(['biological.length.US.filename'], dtype='object')`

But this then requires the same hard-coding to convert from this to `'biological.length.us.filename'` (i.e. `config_identifier`), albeit in simpler fashion:
`config_name = flat_configuration_table.columns[flat_configuration_table.isin([file]).any()]`
`config_name[0].lower()` ---> `'biological.length.us.filename'`

This may become moot w.r.t suggestions related to `L173-179` (i.e. `for data_type in ["biological", "stratification", "NASC", "kriging"]: `).

EchoPro/survey.py
brandynlucca: This happens for cases such as `self.acoustics` that is initialized since it will also store other acoustic information (e.g. `sigma_bs` besides `nasc`). Consequently, the outermost layer of the config tree in the YAML file is `NASC` instead of acoustics. Therefore, it is missing. I suppose this could be initialized via: 
```
self.acoustics = {
    'nasc_df': pd.DataFrame() 
}

self.statistics = {
    'kriging: {
        'mesh_df': pd.DataFrame() ,
        'isobath_200m_df': pd.DataFrame() ,
        'vario_krig_para_df': pd.DataFrame()
    }
}
```

and so on. I'll briefly play around with this, or simply add this structure to the API (probably the easiest option).

leewujung: Might be useful to initialize the dicts with the structure you anticipate, so that the intention is clear, and errors are thrown automatically.

Issue #42, opened 2022-11-17
Bugs in computing sex proportion and length distribution of fish


b-reyes:
To compute the biomass density several constants need to be created. It appears that some of these constants are not correctly created. Specifically, the following should be reviewed: 
- Several proportions and fractions are created in the function [_compute_proportions](https://github.com/uw-echospace/EchoPro/blob/47568e4a96ad74ac1c15167b0e314c8407a09ad3/EchoPro/compute_biomass_density/compute_biomass_density.py#L366-L368) and some of them do not appear to be correct
   - All values that should be investigated are marked with `TODO`
- In the function [_fill_len_wgt_prod](https://github.com/uw-echospace/EchoPro/blob/47568e4a96ad74ac1c15167b0e314c8407a09ad3/EchoPro/compute_biomass_density/compute_biomass_density.py#L491-L492) we are attempting to obtain the length key for station 2 for all fish. However, in the Matlab code (and as a result in the Python code) we exclude the unsexed fish. 
- The function [_get_age2_wgt_prop](https://github.com/uw-echospace/EchoPro/blob/47568e4a96ad74ac1c15167b0e314c8407a09ad3/EchoPro/compute_biomass_density/compute_biomass_density.py#L674-L683) computes the age 2 weight proportion using the specimen data. However, there is a deliberate dropping of NA values that only occurs in this calculation. We should make it so that this is done when the data is first loaded and then use this data for all calculations. 

**@leewujung update [2023/12/05]**:
- `_fill_len_wgt_prod` is now called `_fill_averaged_weight`
- `_get_age2_wgt_prop` is now called `_get_all_age_weight_proportions`
- now sure what age 2 above means -- probably just "adult"

leewujung:
The latest status where `_get_distribution_lengths_station_2` should be corrected:
https://github.com/uw-echospace/EchoPro/blob/2c832292cb9cbae0b5bfd07c0b1ef328043d94c2/EchoPro/computation/transect_results.py#L529-L535

leewujung:
Closing this now that we have traced the matlab code to understand all details and implemented a clean version of python code.

PR #164, opened 2024-01-15
Brandynlucca-WIP-refactor-compute_transect_results

brandynlucca:
Rough draft of the workflow design that is used to convert NASC values into biomass estimates and subsequently a stratified mean/coefficient of variation (CV) for all values (or grouped by specific contrasts, such as age-class or sex).

### DIFF
--- a/EchoPro/computation/operations.py
+++ b/EchoPro/computation/operations.py
@@ -5,9 +5,9 @@ from typing import Union , List
 from typing import Callable
     
 @patch_method_to_DataFrame( pd.DataFrame )
-def discretize_variable( dataframe: pd.DataFrame , 
-                         bin_values: np.ndarray ,
-                         bin_variable: str ):
+def bin_variable( dataframe: pd.DataFrame , 
+                  bin_values: np.ndarray ,
+                  bin_variable: str ):
     """
     Discretizes the target variable into user-defined bins
 
@@ -32,12 +32,12 @@ def discretize_variable( dataframe: pd.DataFrame ,
     )
     
 @patch_method_to_DataFrame( pd.DataFrame )
-def quantize_variables( dataframe: pd.DataFrame ,
-                        bin_variable: str ,
-                        bin_values: np.ndarray,
-                        contrasts: Union[str, List[str]] = [] ,
-                        variables: Union[str, List[str]] = ['length', 'weight'] ,
-                        functions: Union[str, List[str]] = ['mean', 'size'] ):
+def bin_stats( dataframe: pd.DataFrame ,
+               bin_variable: str ,
+               bin_values: np.ndarray,
+               contrasts: Union[str, List[str]] = [] ,
+               variables: Union[str, List[str]] = ['length', 'weight'] ,
+               functions: Union[str, List[str]] = ['mean', 'size'] ):
     """
     Quantizes dataset given user-defined intervals/bins
 
@@ -90,7 +90,7 @@ def quantize_variables( dataframe: pd.DataFrame ,
        
     return (
         dataframe # input dataframe 
-        .discretize_variable( bin_values , bin_variable ) # discretize variable into bins )
+        .bin_variable( bin_values , bin_variable ) # discretize variable into bins )
         .groupby( [f'{bin_variable}_bin'] + con_lst ) # group by these variables/contrasts
         .agg( aggregation_dict ) # apply specified functions
         .replace( np.nan , 0 ) # replace NaN w/ 0's
--- a/EchoPro/computation/operations.py
+++ b/EchoPro/computation/operations.py
@@ -1,0 +1,157 @@
+import numpy as np
+import pandas as pd
+from ..utils.monkey_patch_dataframe import patch_method_to_DataFrame
+from typing import Union , List
+from typing import Callable
+    
+@patch_method_to_DataFrame( pd.DataFrame )
+def discretize_variable( dataframe: pd.DataFrame , 
+                         bin_values: np.ndarray ,
+                         bin_variable: str ):
+    """
+    Discretizes the target variable into user-defined bins
+
+    Parameters
+    ----------
+    dataframe: pd.DataFrame
+        A DataFrame object containing length and age variables required
+        for analyses handling binned variables
+    bin_values: np.ndarray
+        An array of discrete bin values used to construct the intervals for 
+        the variable of interest
+    bin_variable: str
+        The variable that will be binned into discrete intervals
+    Notes
+    -----
+    This will add a column to defined dataframes that groups length and age values
+    into each bin that is explicitly defined in the function arguments
+    """
+    return (
+        dataframe # input dataframe
+        .assign( **{f'{bin_variable}_bin': lambda x: pd.cut(x[bin_variable], bin_values)} ) # assign bin
+    )
+    
+@patch_method_to_DataFrame( pd.DataFrame )
+def quantize_variables( dataframe: pd.DataFrame ,
+                        bin_variable: str ,
+                        bin_values: np.ndarray,
+                        contrasts: Union[str, List[str]] = [] ,
+                        variables: Union[str, List[str]] = ['length', 'weight'] ,
+                        functions: Union[str, List[str]] = ['mean', 'size'] ):
+    """
+    Quantizes dataset given user-defined intervals/bins
+
+    Parameters
+    ----------
+    dataframe: pd.DataFrame
+        A DataFrame object containing length and age variables required
+        for analyses handling binned variables
+    bin_values: np.ndarray
+        An array of discrete bin values used to construct the intervals for 
+        the variable of interest
+    bin_variable: str
+        The variable that will be binned into discrete intervals 
+    contrasts: str or List[str]
+        Additional variables to group data by, such as sex or species
+    variables: str or List[str]
+        Data columns to quantize
+    functions: str or List[str] or List[Callable]
+        Summary statistics or other functions applied to quantized variables
+    """
+    # Ensure that the contrasts, if there are any, are a list in case input is just a str
+    con_lst = [contrasts] if isinstance(contrasts, str) else contrasts
+    
+    # Ensure variables are a list in case input is just a str
+    var_lst = [variables] if isinstance(variables, str) else variables
+    
+    # Ensure functions are contained within a list
+    fun_lst = [functions] if isinstance(functions, (str, Callable)) else functions
+    
+    # Rename functions as needed and if they exist
+    FUNCTION_ALIASES = {
+        'mean': 'mean' ,
+        'size': 'n' ,        
+    }
+    
+    # Check against an allowed list 
+    invalid_functions = set( functions ) - FUNCTION_ALIASES.keys()
+    
+    if invalid_functions:
+        raise ValueError(f'Invalid aggregation functions provided: {invalid_functions}. Only {FUNCTION_ALIASES.keys()} are allowed.')
+
+    # Construction the aggregation dictionary to apply the summary statistics toward
+    aggregation_dict = {
+        variable: [
+            (f"{FUNCTION_ALIASES.get(agg, agg)}_{variable}", agg)
+            for agg in fun_lst
+        ]
+        for variable in var_lst
+    }
+       
+    return (
+        dataframe # input dataframe 
+        .discretize_variable( bin_values , bin_variable ) # discretize variable into bins )
+        .groupby( [f'{bin_variable}_bin'] + con_lst ) # group by these variables/contrasts
+        .agg( aggregation_dict ) # apply specified functions
+        .replace( np.nan , 0 ) # replace NaN w/ 0's
+        .droplevel( level = 0 , axis = 1 ) # drop the column indices 
+        .reset_index( ) # reset the row indices
+    )
+    
+@patch_method_to_DataFrame( pd.DataFrame )
+def count_variable( dataframe: pd.DataFrame ,
+                    contrasts: Union[str, List[str]] ,
+                    variable: str ,
+                    fun: str):
+    """
+    Quantizes dataset given user-defined intervals/bins
+
+    Parameters
+    ----------
+    dataframe: pd.DataFrame
+        A DataFrame object containing length and age variables required
+        for analyses handling binned variables
+    contrasts: str or List[str]
+        Additional variables to group data by, such as sex or species
+    variable: str
+        Data column to bin and count
+    """
+    return (
+        dataframe # input dataframe
+        .reset_index( drop=True )
+        .groupby( contrasts ) 
+        .agg({variable: [('count' , fun)]})
+        .replace(np.nan, 0 )
+        .droplevel( level = 0 , axis = 1 )
+        .reset_index()
+        .sort_values(contrasts)
+    ) 
+    
+@patch_method_to_DataFrame( pd.DataFrame )
+def meld( specimen_dataframe: pd.DataFrame ,
+          length_dataframe: pd.DataFrame ):
+    """
+    Concatenates the specimen and length dataframes using a shared format
+
+    Parameters
+    ----------
+    specimen_dataframe: pd.DataFrame
+        A DataFrame object containing data from the specimen dataset
+    length_dataframe: pd.DataFrame
+        A DataFrame object containing data from the length dataset
+    """
+    # Reorganize the specimen dataframe so it matches the format
+    # of the length dataframe w/ length counts
+    specimen_stacked = (
+        specimen_dataframe 
+        .copy()
+        .groupby(['stratum_num' , 'species_id' , 'sex' , 'group' , 'station' , 'length' , 'length_bin' ])
+        .apply(lambda x: len(x['length']))
+        .reset_index(name='length_count')
+    )
+    
+    # Concatenate the data frames and return
+    return pd.concat( [ specimen_stacked ,
+                        length_dataframe ] ,
+                        join = 'inner' )
+    
--- a/EchoPro/survey.py
+++ b/EchoPro/survey.py
@@ -131,10 +131,8 @@ class Survey:
         
         # Amend length/age distribution locations within the configuration attribute
         config_to_add[ 'biometrics' ] = { 
-                                         'parameters': {
-                                            'bio_hake_len_bin': init_config_params[ 'bio_hake_len_bin' ] ,
-                                            'bio_hake_age_bin': init_config_params[ 'bio_hake_age_bin' ]
-                                         } 
+            'bio_hake_len_bin': init_config_params[ 'bio_hake_len_bin' ] ,
+            'bio_hake_age_bin': init_config_params[ 'bio_hake_age_bin' ]
         }
         
         del config_to_add['bio_hake_len_bin'] , config_to_add['bio_hake_age_bin']
@@ -346,14 +344,14 @@ class Survey:
         """
 
         # Pull the relevant age and length bins and output a dictionary
-        length_bins = np.linspace( self.config['bio_hake_len_bin'][0] ,
-                           self.config['bio_hake_len_bin'][1] ,
-                           self.config['bio_hake_len_bin'][2] ,
+        length_bins = np.linspace( self.config[ 'biometrics' ]['bio_hake_len_bin'][0] ,
+                           self.config[ 'biometrics' ]['bio_hake_len_bin'][1] ,
+                           self.config[ 'biometrics' ]['bio_hake_len_bin'][2] ,
                            dtype = np.float64 )
 
-        age_bins = np.linspace( self.config['bio_hake_age_bin'][0] ,
-                                self.config['bio_hake_age_bin'][1] , 
-                                self.config['bio_hake_age_bin'][2] ,
+        age_bins = np.linspace( self.config[ 'biometrics' ]['bio_hake_age_bin'][0] ,
+                                self.config[ 'biometrics' ]['bio_hake_age_bin'][1] , 
+                                self.config[ 'biometrics' ]['bio_hake_age_bin'][2] ,
                                 dtype = np.float64 )
 
         ### Discretize the age and length arrays into user-defined bins that will be used later on
@@ -504,16 +502,16 @@ class Survey:
             specimen_df_copy
             .groupby(['haul_num', 'stratum_num' , 'species_id', 'length'])
             .apply(lambda x: len(x['length']))
-            .reset_index(name='length_count')
+            .reset_index(name= 'length_count' )
             )
         
         ### Concatenate the two dataframes
         all_length_df = pd.concat( [ spec_df_reframed , length_df_copy ] , join = 'inner' )
         
         # Import parameters from configuration
-        ts_length_parameters = self.config['TS_length_regression_parameters']['pacific_hake']
-        slope = ts_length_parameters['TS_L_slope']
-        intercept = ts_length_parameters['TS_L_intercept']
+        ts_length_parameters = self.config[ 'TS_length_regression_parameters' ]['pacific_hake']
+        slope = ts_length_parameters[ 'TS_L_slope' ]
+        intercept = ts_length_parameters[ 'TS_L_intercept' ]
         
         # Convert length values into TS
         ### ??? TODO: Not necessary for this operation, but may be useful to store for future use ?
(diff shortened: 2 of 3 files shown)


EchoPro/survey.py
```diff
@@ -363,6 +364,155 @@ def biometric_distributions( self ):
...
+        # Calculate mean value per stratum here 
+        
+    def impute_missing_strata():
+        
+    def fit_length_weight_relationship( length_df , specimen_df ):
+        
+    
+        
+    @staticmethod
+    def TS_L_regression( length_df , specimen_df ):
```
leewujung, line 487: I know this is just a sketch, so just a small note that if this is not a static method then `length_df` and `specimen_df` are accessible through `self`.

leewujung (commented):
Hey @brandynlucca : Thanks for sending the design! I am SO looking forward to having these clean functions and organized dataframes/parameters. :) 

I have some small inline comments but I like the breakdown you have. The only thing I would suggest on a higher level is how input data and computation outputs are stored in the `Survey` object.

What do you think to have a clear separation between the inputs and the outputs?

Say for the input dataframes, change what you have in [here](https://github.com/uw-echospace/EchoPro/blob/81dd42bd8f6e0b92938ccf2bfd107f4549226cb9/EchoPro/survey.py#L58-L62) to
```python
# Initialize data attributes ! 
self.input["acoustics"] = copy.deepcopy(LAYER_NAME_MAP['NASC']['data_tree'])
self.input["biology"] = copy.deepcopy(LAYER_NAME_MAP['biological']['data_tree'])
self.input["spatial"] = copy.deepcopy(LAYER_NAME_MAP['stratification']['data_tree'])
self.input["statistics"] = copy.deepcopy(LAYER_NAME_MAP['kriging']['data_tree'])
```

For the computed output, gathering what's currently in the sketch (and fully recognizing it that it is just a sketch that has not been finalized):
```python
# derived values from inputs
self.acoustics['sigma_df'] : pd.DataFrame
self.statistics['length_weight_arr'] : np.ndarray
self.biology['weight_df'] : pd.DataFrame

self.biology['population']['areal_density_df'] : pd.DataFrame
self.biology['population']['abundance_df'] : pd.DataFrame
self.biology['population']['biomass_df'] : pd.DataFrame

self.results['transect_results'] : dict
self.results['transect_results']: {'mean': np.float64, 'cv': np.float64}

# kriging
self.statistics['kriging']['vario_krig_para_df']
self.statistics['semivariogram'] : dict
self.statistics['kriging']['modeled_biomass'] : pd.DataFrame
self.results['kriging_results'] : dict
self.results['kriging_results']: {'mean': np.float64, 'cv': np.float64}
```

What do you think to just have 2 output attributes `self.input_derived` and `self.kriging`:
```python
# Input-derived values
# --- from hauls
self.input_derived['sigma_df'] : pd.DataFrame
self.input_derived['length_weight_arr'] : np.ndarray
self.input_derived['weight_df'] : pd.DataFrame

# --- from combining haul-drived results with NASC
self.input_derived['areal_density_df'] : pd.DataFrame
self.input_derived['abundance_df'] : pd.DataFrame
self.input_derived['biomass_df'] : pd.DataFrame

# --- from  Jolly-Hampton stratified analysis
self.input_derived['jolly_hampton_mean']: np.float64
self.input_derived['jolly_hampton_cv']: np.float64


# kriging
# --- from pre-set params
self.kriging['vario_krig_para_df'] : pd.DataFrame
self.kriging['semivariogram_params'] : dict

# --- this is what you have right now
self.kriging['modeled_biomass'] : pd.DataFrame
# --- I wonder if we can use the same name as the input-derived ones
self.kriging['areal_density_df'] : pd.DataFrame
self.kriging['abundance_df'] : pd.DataFrame
self.kriging['biomass_df'] : pd.DataFrame

# --- from  Jolly-Hampton stratified analysis
self.kriging['jolly_hampton_mean']: np.float64
self.kriging['jolly_hampton_cv']: np.float64
```

My rationale is that this provides the user a clear expectation/view of where they are in the process: input --> values derived from inputs --> values after kriging

leewujung:
Separately, for this section below
https://github.com/uw-echospace/EchoPro/blob/81dd42bd8f6e0b92938ccf2bfd107f4549226cb9/EchoPro/survey.py#L58-L62
I wonder why you didn't want to use the same name as in the LAYER_NAME_MAP dict, like:
```python
        # Initialize data attributes ! 
        self.nasc = copy.deepcopy(LAYER_NAME_MAP['NASC']['data_tree'])
        self.biological = copy.deepcopy(LAYER_NAME_MAP['biological']['data_tree'])
        self.stratification = copy.deepcopy(LAYER_NAME_MAP['stratification']['data_tree'])
        self.kriging = copy.deepcopy(LAYER_NAME_MAP['kriging']['data_tree'])
```
Then you could actually do something like
```python
for data_type in ["NASC", "biological", "stratification", "kriging"]
   setattr(self, data_type, copy.deepcopy(LAYER_NAME_MAP[data_type]["data_tree"]))
```
And if the acoustics/biology/spatial/statistics names are better, you could change those in `LAYER_NAME_MAP` to achieve the same thing -- though knowing how locked in people are talking about stratification and NASC etc, it may be good to keep the original names, since it is easier for the users.

EchoPro/survey.py
```diff
@@ -363,6 +364,155 @@ def biometric_distributions( self ):
...
+        
+        # Apply length-weight regression to biological data
+        ### This will largely resemble the sigma_bs calculation
+        ### since it also follows the "use regression to calculate value"
+        ### workflow. So it may be more parsimonious to bundle these 
+        ### under a shared regression function
+        #### This also subsume the original '_get_weight_num_fraction_adult'
+        #### INPUT: 'group' = 'all_ages', 'adult'
+        self.strata_sex_weight_all( ... , group = ...)
+        # OUTPUT: self.biology['weight_df'] (pd.DataFrame)
```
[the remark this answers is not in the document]
brandynlucca: I totally agree. I will leave this conversation as unresolved for reference when I fully flesh out these two functions.

EchoPro/survey.py
```diff
@@ -363,6 +364,155 @@ def biometric_distributions( self ):
...
+        ### It also does not need to resemble the original source code (ie for loop)
+        missing_strata = []
+        for stratum in self.spatial['strata_df']['stratum_num']:
+            if stratum not in self.acoustics['sigma_df']['stratum_num']:
+                missing_strata.append(stratum)
+        # OR 
+        self.impute_missing_strata()
+        
+        # Fill in missing sigma_bs values
+        self.impute_missing_sigma_bs()
```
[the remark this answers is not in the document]
brandynlucca: I think that is certainly workable. Would it be too much to embed both `impute_missing_strata()` and `impute_missing_sigma_bs()` within `strata_mean_sigma_bs()`? I can see how that may obfuscate the fact that the underlying (modeled) data are being imputed and therefore be a tad more difficult to debug/parse.

brandynlucca:
Thanks for the feedback @leewujung! Along with replying to specific in-line comments, I have included below a few replies to your overall comments about the object structure. 

> What do you think to have a clear separation between the inputs and the outputs?

I see the utility of this, particularly when differentiating the 'raw' data (i.e. data output from the initial loader). I think this will become clearer as the remaining code is fleshed out, but I suppose it may be worth adding a referential data processing label to new attributes as I develop things. For instance (and not exhaustive):

* **L0 (pre-processed data)**    
    * YAML config files
    * Unprocessed data input files  
* **L1 (`Survey` loaded data)**
   * Biological
   * Kriging
   * NASC
   * Stratification
* **L2 (Derived/intermediate variables required for later calculations)**
   * Regression outputs (sigma_bs, weight)
   * Statistical distributions (length, age)
   * Kriging parameterization/fitting
   * Semivariogram model parameterization/fitting
* **L3 (georeferenced data products)**
   * Areal density
   * Abundance
   * Biomass
* **L4 (summary data products)**
   * Stratified mean (transect, kriging, and a combination of these w/ bootstrapping)
   * Stratified variance/coefficient of variation (transect, kriging, and a combination of these w/ bootstrapping)

Or something to that effect.

> And if the acoustics/biology/spatial/statistics names are better, you could change those in `LAYER_NAME_MAP` to achieve the  same thing -- though knowing how locked in people are talking about stratification and NASC etc, it may be good to keep the original names, since it is easier for the users.

The rationale for using `acoustics` over `nasc` was to allow for organizing all of the acoustic attributes into a catch-all category, namely any `sigma_bs` intermediate variables that bridge `nasc` to the biologically meaningful parameters (e.g. biomass). This was also the case for `statistics` since notebooks included examples of stratified summary statistics being calculated for both the transect data, kriging-based estimates, and some combination of bootstrapping, etc. I think this was also from before I had reduced the generalizability (via the recursive functions) where I had other user inputs in mind. I think I elected for `spatial` since that seemed more intuitive for me, but I agree may be better left as `stratification` (although I am strangely biased since I find `spatial` much easier to type out than `stratification`... weird). If the original variable names are preferred, then that should be a simple enough change to implement.

EchoPro/survey.py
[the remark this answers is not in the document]
leewujung: I agree that it would be ok to embed both  `impute_missing_strata()` and `impute_missing_sigma_bs()` within `strata_mean_sigma_bs()`, so that `strata_mean_sigma_bs` will output `sigma_bs` for each strata, regardless of what happens under the hood. The imputation seems a bit too detailed to split out, but is something that could/should be explained in the docstring, so that users are reminded that this may have happened to their data.

Actually, how about if you indeed embed everything under `strata_mean_sigma_bs` and print out messages like "Length data missing for stratum A, using strata B-C for estimating sigma_bs for stratum A" for each of the missing strata?

brandynlucca:
[Commit #828adb](https://github.com/uw-echospace/EchoPro/pull/164/commits/828adb21e2c14fde30f6cf2ad1916251b49b312a)  includes some modifications to how certain data outputs are formatted so they are compatible for later use (when needed). This also includes processing steps through calculating the sex-specific proportions and mean strata weights. 

There is a minor discrepancy in this calculation that provides different values (on the order of 0.001 - 0.01) for the stratum-specific average weights than some of the ones estimated from the previous code stored under `self.bio_calc.bio_param_df`. I have added a separate comment pointing out this code block.

EchoPro/survey.py
```diff
@@ -562,98 +593,237 @@ def impute_missing_sigma_bs(self):
...
+        station_sex_length = (
+            specimen_grouped # begin reformatting specimen_grouped so it resembles same structure as length_grouped
+            .meld( length_grouped ) # "meld": reformats specimen_grouped and then concatenates length_grouped 
+            # sum up all of the length values with respect to each length bin
+            .count_variable( contrasts = [ 'group' , 'station' , 'stratum_num' , 'length_bin' ] , # grouping variables
+                             variable = 'length_count' , # target value to apply a function
+                             fun = 'sum' ) # function to apply
+        )
+        
+        ### Calculate total sample size ('group' == 'all') that will convert 'counts' into 'frequency/proportion'
```
brandynlucca, line 702: The discrepancy between the original code and how this is summed here. This seems to be how the stratum sample size was calculated; the original code expressly omits values where `sex = 3` from the `specimen` dataset. There was a TODO comment included in there that just used the full sum of both the `length` and `specimen` dataframes for that calculation. 

For the most part, these differences were non-existence where `sex=3` was a trivial fraction of the data. When `stratum=2`, this is where the largest differences are/were observed, which coincides with a non-trivial number of unsexed animals where both the male and female samples comprise ~ 25% of the overall sample (i.e. ~ 50% of the sample from both datasets were sexed). So while the proportions for both `male` and `female` are identical between the original and updated code, the mean weights are slightly different.

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+
+        # Now the bins are centered with the first and last elements properly appended
+        # These create an along-array interval such that values can be cut/discretized into
+        # bins that fall between each value/element of the array
+        length_centered_bins = np.concatenate( ( [ length_bins[0] - length_binwidth ] ,
+                                                length_bins + length_binwidth ) )
+        age_centered_bins = np.concatenate( ( [ age_bins[0] - age_binwidth ] ,
+                                            age_bins + age_binwidth ) ) 
+
+        # Construct dictionary that will be move values within the configuration file
```
leewujung, line 361: Some words are missing or there's a typo of some kind in this sentence?

leewujung: Maybe as a result, I am not understanding why the same parameters are saved again in a different place in `self.config`. Perhaps `self.config['bio_hake_len_bin']` should just be saved in `self.config['biometrics']['bio_hake_len_bin']` when it's first parsed and saved?

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+        Calculates the stratified mean sigma_bs for each stratum
+
+        Parameters
+        ----------
+        species_id : np.float64
+            Numeric code representing a particular species of interest
+
+        Notes
+        -----
+        This function iterates through eahc stratum to fit acoustic target
```
leewujung, line 483: Just fixing a typo:
```suggestion
        This function iterates through each stratum to fit acoustic target 
```

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+            
+            # Find strata intervals to impute over        
+            for i in missing_strata:
+                strata_floor = present_strata[present_strata < i]
+                strata_ceil = present_strata[present_strata > i]
+
+                new_stratum_below = np.max(strata_floor) if strata_floor.size > 0 else None
+                new_stratum_above = np.min(strata_ceil) if strata_ceil.size > 0 else None      
+                
+                sigma_bs_indexed = sigma_bs_impute[sigma_bs_impute['stratum_num'].isin([new_stratum_below, new_stratum_above])]
```
leewujung, line 592: Not sure if I get this right, but seems that if `new_stratum_below` and `new_stratum_above` are not off by more than 1, the statement `.isin([new_stratum_below, new_stratum_above])]` would have some error.

Say if `new_stratum_below=2` and `new_stratum_above=5`, `.isin` would not get everything between 2 and 5.

leewujung: This calls for a separate cases in the same test function to make sure the imputation is correct, something like:
Case 1: new_stratum_below=2, new_stratum_above=4
Case 2: new_stratum_below=2, new_stratum_above=5

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+        # Calculate the weighted mean sigma_bs per haul
+        ### This will track both the mean sigma_bs and sample size since this will propagate as a
+        ### grouped mean contained with a shared stratum
+        mean_haul_sigma_bs = (
+            all_length_df
+            .groupby(['haul_num' , 'stratum_num' , 'species_id' ])[['sigma_bs' , 'length_count']]
+            .apply(lambda x: np.average( x['sigma_bs'] , weights=x['length_count']))
+            .to_frame('mean_sigma_bs')
+            .reset_index()
+        )
```
leewujung, line 531: I like how clean this is!

EchoPro/computation/operations.py
```diff
@@ -0,0 +1,157 @@
+import numpy as np
+import pandas as pd
+from ..utils.monkey_patch_dataframe import patch_method_to_DataFrame
+from typing import Union , List
+from typing import Callable
+    
+@patch_method_to_DataFrame( pd.DataFrame )
+def discretize_variable( dataframe: pd.DataFrame , 
+                         bin_values: np.ndarray ,
+                         bin_variable: str ):
```
leewujung, line 10: Similar to the case with `quantize_variables`, I think `discretize_variable` may be better named as `bin_variable`

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+        }
+        
+        # summarize mean/samples of length and weight values (binned statistics)
+        length_bin_stats = length_weight_df.quantize_variables( bin_variable = 'length' ,
+                                                                bin_values = length_intervals )
+        
+        # fill bins where n_length < 5 w/ regressed weight values
+        length_bin_stats['modeled_weight'] = length_bin_stats['mean_weight'].copy()
+        low_n_indices = np.where(length_bin_stats['n_weight'].values < 5)[0].copy()   
+        length_bin_stats.loc[low_n_indices, 'modeled_weight'] = fitted_weight[low_n_indices].copy()
```
leewujung, line 644: This is just a style thing:
If I remember correctly, echopro variable names have been following the pattern of "noun_modifier", so `weight_mean`, `weight_n`, `weight_modeled` would better conform with that convention.

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+        ----------
+        species_id : np.float64
+            Numeric code representing a particular species of interest
+
+        Notes
+        -----
+        This function first fits a length-weight regression based on measured 
+        values and then produces an array of fitted weight values (based on 
+        binned length values) that are used for later biomass calculations, 
+        weighting, and apportionment.
```
leewujung, line 614: Minor edits to improve clarity.
```suggestion
        Notes
        -----
        This function first fits a length-weight regression based on measured 
        values and then produces an array of fitted weight values based on 
        binned length values.
        The length-weight relationship produced here are used for later 
        biomass calculations and apportionment.
```

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+            .loc[ station_length_aggregate.group.isin( ['male' , 'female'] ) ] # only parse 'male' and 'female'
+            # create a pivot that will reorient data to the desired shape
+            .pivot_table( index = [ 'group' , 'station' ] , 
+                          columns = [ 'stratum_num' ] , 
+                          values = [ 'p_overall_station' ] )
+            .groupby( 'group' )
+            .sum()
+        )
+        
+        ### Calculate the the proportions each dataset / station contributed within each stratum
```
leewujung, line 737: remove the extra the
```suggestion
        ### Calculate the proportions each dataset / station contributed within each stratum
```

EchoPro/survey.py
```diff
@@ -327,269 +328,502 @@ def read_validated_data( self ,
...
+        
+        ### Store the data frame in an accessible location
+        self.biology[ 'weight' ][ 'weight_strata_df' ] = pd.DataFrame( {
+            'stratum': total_weight.index ,
+            'female_proportion': sex_proportions.loc[ 'female' , : ][ 'p_overall_station' ].reset_index(drop=True) ,
+            'male_proportion': sex_proportions.loc[ 'male' , : ][ 'p_overall_station' ].reset_index(drop=True) ,
+            'female_average_weight': female_weight ,            
+            'male_average_weight': male_weight ,
+            'total_average_weight': total_weight
+        } )
```
leewujung, line 821: Following the above "noun_modifier" comment, maybe these become `proportion_male/female` and `average_weight_male/female/total`?

leewujung (commented):
Hey @brandynlucca : Thanks for pushing the changes! I have some small inline comments. I think once those are resolved, you can go ahead to merge this PR.

Apart from the inline comments, I have 2 suggestions:

The use of `patch_method_to_DataFrame` seems to be doing what `.pipe` intends to do (or probably more aking to the [accessors](https://pandas.pydata.org/docs/development/extending.html#registering-custom-accessors)). Since you already use `.pipe` in various places, I'd suggest going the `.pipe` route and not using the decorator/accessor to keep the code easily traceable. 
 Since your functions are all working fine now, I think we can just keep them as is. I created #166 so that we keep track of that after this main push.

The other is that I think this is time to set up pre-commit so that the code is automatically formatter. The lines in `strata_sex_weight_proportions` in particular are very long. I suggest doing a one-time commit in a separate PR to push up those auto-formatted changes at some point soon, probably after the kriging overhaul. I created #167 for that.

leewujung:
Separate from my comments above: 

I have to admit that I did not go through all the details in `strata_sex_weight_proportions` (thanks for wading through everything...!). How does the new implementation change the status of the problematic matlab/previous implementation documented in #42? I assume that you implemented what the correct operation should be?

EchoPro/survey.py
brandynlucca: For reference, this is what `self.acoustics['sigma_bs']['strata_mean']` should look like when no strata are missing: 

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000425
1          1.0       22500       0.000084
2          2.0       22500       0.000070
3          3.0       22500       0.000114
4          4.0       22500       0.000191
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
8          8.0       22500       0.000481
```

**Case 1**

Dropping `stratum_num = 3` via:

```python
strata_mean_test = strata_mean.loc[ strata_mean[ 'stratum_num' ] != 3 ]
``` 

yields for `sigma_bs_impute`:

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000425
1          1.0       22500       0.000084
2          2.0       22500       0.000070
0          3.0       22500            NaN
4          4.0       22500       0.000191
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
8          8.0       22500       0.000481
```

When running the loop nested within `for i in missing_strata: ...`, this then yields: 

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000425
1          1.0       22500       0.000084
2          2.0       22500       0.000070
0          3.0       22500       0.000130
4          4.0       22500       0.000191
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
8          8.0       22500       0.000481
```

which produces the expected result (i.e. the mean of strata 2 and 5). 

**Case 2**

Dropping `stratum_num = [ 3 , 4]` via:

```python
strata_mean_test = strata_mean.loc[ ~ strata_mean[ 'stratum_num' ].isin( [ 3 , 4 ] ) ]
```

does trigger an error when generating `sigma_bs_impute` due to the mismatch in the number of elements for each column. So this has been modified in the updated commit to correctly produce the following dataframe: 

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000425
1          1.0       22500       0.000084
2          2.0       22500       0.000070
0          3.0       22500            NaN
1          4.0       22500            NaN
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
8          8.0       22500       0.000481
```

When run through the rest of the function, this then yields the expected output where the values for strata 3 and 4 are filled with the averaged `mean_sigma_bs` from strata 2 and 5: 

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000425
1          1.0       22500       0.000084
2          2.0       22500       0.000070
0          3.0       22500       0.000183
1          4.0       22500       0.000183
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
8          8.0       22500       0.000481
```

**Additional cases**

When testing for the edge cases where the extreme strata values are missing via:

```python
strata_mean_test = strata_mean.loc[ ~ strata_mean[ 'stratum_num' ].isin( [ 0 , 8 ] ) ]
```
```python
0          0.0       22500            NaN
1          1.0       22500       0.000084
2          2.0       22500       0.000070
3          3.0       22500       0.000114
4          4.0       22500       0.000191
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
1          8.0       22500            NaN
```

The function then correctly imputes the nearest-neighbor values:

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000084
1          1.0       22500       0.000084
2          2.0       22500       0.000070
3          3.0       22500       0.000114
4          4.0       22500       0.000191
5          5.0       22500       0.000295
6          6.0       22500       0.000247
7          7.0       22500       0.000384
1          8.0       22500       0.000384
```

In a messier case where the edge cases are missing, a single stratum is missing from the dataset, and then back-to-back strata are also missing from the dataset:

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500            NaN
1          1.0       22500       0.000084
2          2.0       22500       0.000070
1          3.0       22500            NaN
4          4.0       22500       0.000191
2          5.0       22500            NaN
3          6.0       22500            NaN
7          7.0       22500       0.000384
4          8.0       22500            NaN
```

The function appears to impute the mean/nearest-neighbor `mean_sigma_bs` for the missing values: 

```python
   stratum_num  species_id  mean_sigma_bs
0          0.0       22500       0.000084
1          1.0       22500       0.000084
2          2.0       22500       0.000070
1          3.0       22500       0.000130
4          4.0       22500       0.000191
2          5.0       22500       0.000287
3          6.0       22500       0.000287
7          7.0       22500       0.000384
4          8.0       22500       0.000384
```

So this otherwise appears to function as intended when fixing the concatenation of the imputation dataframe.

PR #176, opened 2024-02-08
Add population-level calculations and stratified statistics

brandynlucca:
Several functions were added to calculate the age- and sex-stratified areal number/biomass densities, summed abundance, and summed biomass estimates. This includes the following changes:

- `initialization_config.yml`: added a field for `replicates` specific to the stratified resampling procedure.
- `spatial.py`
   - `calculate_bounds`: calculates the latitude/longitude boundary box around each transect line
   - `calculate_transect_distance`: calculates the length distance and area of each transect line necessary for the stratified statistic calculation
-  `statistics.py`
   - `confidence_interval`: calculates the 95% confidence interval (for a Normal distribution)
   - `stratified_transect_statistic`: produces estimates and confidence intervals (around the mean) for the following statistics/estimators: mean, variance, coefficient of variation (CV)
- `survey.py`
   - `nasc_to_biomass_conversion`: converts NASC and biological data into areal biomass/number density, summed biomass, and summed abundance stratified by age, sex, stratum, and transect
      - Nested within `transect_analysis` (pre-existing `Survey` method)
   - `stratified_summary`: wrapper function for `stratified_transect_statistic` and `calculate_transect_distance` that adds mean and confidence interval estimates for the stratified mean, variance, and coefficient of variation (CV) produced by the resampling/bootstrapping procedure
      - New `Survey` method
      
_Current use_

```python
from EchoPro.survey import Survey

# Load configuration settings and all data
obj = Survey( init_config_path='./config_files/initialization_config.yml' , survey_year_config_path='./config_files/survey_year_2019_config.yml' )

# Calculate base transect-based results necessary for later calculations
obj.transect_analysis()

# Calculate transect-based stratified results
obj.stratified_summary()

# Calculate biomass 
print(f"Total adult biomass: {1e-6*self.biology[ 'population' ][ 'biomass' ][ 'biomass_df' ].pipe( lambda df: df.loc[ ( df[ 'sex' ] == 'total' ) ] ).B_adult.sum():.3f} kmt")
```
```
Total adult biomass: 1655.225 kmt
```
```python
print(f"Adult biomass CV: {self.biology[ 'population' ][ 'stratified_results' ][ 'biomass' ][ 'CV' ][ 'estimate' ]*1e2:.2f}% [{ci_bounds[0]*1e2:.2f}%, {ci_bounds[1]*1e2:.2f}%; 95 pct. CI]" )
```
```
Adult biomass CV: 13.32% [12.40%, 14.23%; 95 pct. CI]
```


brandynlucca:
Commit [https://github.com/uw-echospace/EchoPro/pull/176/commits/1cdcf4d6f6117b08b317a0e7775fda9b524dabdd](https://github.com/uw-echospace/EchoPro/pull/176/commits/1cdcf4d6f6117b08b317a0e7775fda9b524dabdd) was also pushed to see how how the CI testing handled the change. The change in `load_configuration` from a class method to a standalone function results in an expected failed test state. This makes sense given that [L22-25](https://github.com/uw-echospace/EchoPro/blob/f79b6d5bba66ed731b8c414cba01b89230c116e5/EchoPro/tests/test_data_loader.py#L22) within `test_data_loader.py` still expects `load_configuration` to be a `Survey` class method. 

I am just noting this as a reminder to myself to amend the `test_data_loader.py` accordingly and have created Issue #177 to track both this failed test and future bugs/errors for the integration tests as a whole.

EchoPro/computation/spatial.py
```diff
@@ -0,0 +1,61 @@
+import pandas as pd
+import geopy.distance
+
+def calculate_bounds( group ,
+                      contrast ):
+    """
+    Calculates latitude/longitude boundary box
```
leewujung, line 7: It is not a box but a line?

EchoPro/tests/test_data_loader.py
```diff
@@ -1,6 +1,7 @@
 import yaml
 from pathlib import Path
 from EchoPro.survey import Survey
```
leewujung, line 3: Remove unused Survey
```suggestion

```

EchoPro/survey.py
```diff
@@ -315,10 +258,17 @@ def read_validated_data( self ,
...
             elif config_map[0] == 'kriging':
                 internal = internal['kriging']    
-            
+
             # A single dataframe per entry is expected, so no other fancy operations are needed
-            df_list = [internal[config_map[1] + '_df'] , df]
-            internal[config_map[1] + '_df'] = pd.concat(df_list)
+            # If geo_strata, differentiate between inpfc and stratification1 sheets
+            ### TODO: Temporary approach for incorporating the inpfc dataset. An improved approach
+            ### can be incorporated later on.
```
leewujung, line 265: Suggest adding this as an issue for tracking.

EchoPro/survey.py
```diff
@@ -944,10 +840,287 @@ def strata_age_binned_weight_proportions( self ,
...
+            .loc[ : , [ 'latitude' , 'longitude' , 'transect_num' , 'stratum_num' , 'haul_num' , 'interval_area' , 'NASC_all_ages' , 'NASC_no_age1' ] ]
+        )
+
+        ### Call additional dataframes needed to merge with the NASC data and subsequently calculate
+        ### population-level metrics (and later statistics)
+        # Sex-stratum-indexed proportions and average weight
+        weight_sex_strata = (
+            self.biology[ 'weight' ][ 'weight_strata_df' ]
+            .copy()
+            .rename( columns = { 'stratum': 'stratum_num' } )
```
leewujung, line 890: I wonder if this renaming means that we could rename this column in the data loader.

EchoPro/survey.py
```diff
@@ -944,10 +840,287 @@ def strata_age_binned_weight_proportions( self ,
...
+        nasc_adult_number_proportions = (
+            self.biology[ 'weight' ][ 'age_stratified' ][ 'age_1_excluded' ][ 'number_proportions' ]
+            .rename( columns = { 'number_proportion': 'adult_number_proportion' } )
+        )
+
+        # Adult NASC proportions for each stratum (weight)
+        nasc_adult_weight_proportions = (
+            self.biology[ 'weight' ][ 'age_stratified' ][ 'age_1_excluded' ][ 'weight_proportions' ]
+            .rename( columns = { 'weight_proportion': 'adult_weight_proportion' } )
+        )
```
leewujung, line 915: Same as the above, wondering if this renaming means that we could rename this column in the data loader.

leewujung (changes requested):
@brandynlucca : Thanks for the PR, nice work!

My inline comments are small. Below are some high-level organization comments, regarding the function `nasc_to_biomass_conversion`.

### Some observations
1. L862-L937, the code does 2 things:
   - prepare `dataframes_to_add`, which is used to calculate `nasc_fraction_total_df`
   - prepare `sex_indexed_weight_proportions`, which is used to calculate `age_biomass_df`
2. L939-L950:
   - calculate `nasc_fraction_total_df`
3. L939-L1022:
   - everything is derived from `nasc_to_areal_number_density_df` (and other stored data within the Survey object)
4. L1032-1038:
   - `age_biomass_df` is calculated
5. At the end the key dataframes are saved


### Suggestions
Since this function is very long, I suggest factoring out the section (1. and 2. above) into two separate functions, one that produces `nasc_fraction_total_df` and the other produces `sex_indexed_weight_proportions`. These two functions would be basically helper function that prepares data.

This way the main flow of your calculation will be very clearer, since people who look into the code do not have to keep track of the large number of dfs from the beginning, since they are merged/grouped nicely into columns with good names in `nasc_fraction_total_df` and `sex_indexed_weight_proportions`.

The separation will also help with writing tests, so that if something fails, the chunk to isolate to figure out potential bugs is smaller.

EchoPro/survey.py
brandynlucca: This has been added as an issue (#178).

EchoPro/survey.py
brandynlucca: I think there are two ways to go about this: 

1. I could not be lazy and accidentally name the data column `stratum` instead of `stratum_num`.
2. We can drop the appended `_num` in the data loader so it is consistent with other variables (e.g. `length` instead of `length_num`). That may be a poor example since, as far as I understand, `*_num` indicates the numbered transect, haul, and stratum, and therefore acts as a flag for "hey, this is an indexing variable we care about".

It's not a particularly onerous task to type out '_num', and appending that text also is (probably) never the primary contributor toward very long lines of code that need to be better formatted/broken up. If `*_num` is, in fact, a flag (of sorts) of an ordinal numbered variable, then it may be useful to devise a data dictionary/glossary that at least defines its meaning. I think this also more generally applies to where we append `*_df`, `*_gdf`, `*_arr`, and other similar strings. It is fairly intuitive from our end, but it may not be for folks interacting with the `Survey` objects. This already exists to some degree here:

https://github.com/uw-echospace/EchoPro/blob/f653ba74f445ad97c365bfc077bde602a3b8ab8b/docs/classes-datastructures.md?plain=1#L25

EchoPro/survey.py
brandynlucca: In these examples, I may need to rethink how I am originally generating the dataframes when calculating the sex/age proportions, or just not rename them at all in this code block. In my mind, I assume that someone accessing these age-specific trees understand that the associated `weight_proportion` and `number_proportion` are subsequently adult-only variables. I renamed the columns here since this function could be applied to either all age-classes, just YOY, or something else entirely. I'll take a step back and walk through it again to see where I can trim things, but I think your suggestion regarding further refactoring of the `nasc_to_biomass` function may be the simplest solution here.

EchoPro/computation/spatial.py
brandynlucca: Addressed in Commit [22bf7e1](https://github.com/uw-echospace/EchoPro/pull/176/commits/22bf7e14e654624647a38f26d0749feeff7013a5).

brandynlucca:
@leewujung  I think I have addressed everything you've mentioned so far. I am now in the process of stepping backward to enable the age-1 inclusion/exclusion flag. I can include that in a separate PR while also constructing more basic testing functions.

leewujung:
@brandynlucca : Could you fix test breaking? Seems it's a missing import. Thanks.

leewujung:
> I am now in the process of stepping backward to enable the age-1 inclusion/exclusion flag. I can include that in a separate PR while also constructing more basic testing functions.

I would suggest just having the tests of what you have in a single PR.

Let's do the age-1 inclusion/exclusion later, after kriging and reports/figures, so we have the entire process completed and verified first.

EchoPro/computation/spatial.py
leewujung: Oh actually my comment was regarding the docstring, not the function name.
```suggestion
    Calculates start and end latitude/longitude
```
You may want to make similar changes to the Notes.

EchoPro/computation/statistics.py
```diff
@@ -0,0 +1,128 @@
+import numpy as np
+import scipy.stats as st
+
+def stratified_transect_statistic( transects , 
+                                   strata , 
+                                   sample_fraction , 
+                                   replicates ):
```
leewujung, line 7: How about changing `replicates` to be `num_of_replicates` to be exact?

leewujung: perhaps then the name in the config yml can be changed to.

EchoPro/computation/statistics.py
```diff
@@ -0,0 +1,128 @@
...
+
+    Notes
+    -----
+    This function calculates the stratified summary statistics for biomass within 
+    `EchoPro.survey.stratified_summary()`.
+    """ 
+    
+    ### Convert specific DataFrame columns to arrays for speed
+    distance = transects[ 'transect_distance' ].values
+    value = transects[ 'B_adult' ].values
```
leewujung, line 33: It seems that here the `B_adult` could be swapped out with other values too and the calculation is general. If that is true, how about adding an argument `param_name` and use `param_name="B_adult"` for the calculation?

EchoPro/computation/statistics.py
```diff
@@ -0,0 +1,128 @@
...
+                                         replace=False )
+            
+            ### Define stratified weights
+            stratified_weights = distance[ sel_inds ] / np.mean( distance[ sel_inds ] )
+
+            ### Weighted value (e.g. biomass)
+            value_distance_density = value[ sel_inds ] / distance[ sel_inds ]
+
+            ### Compute mean and variance
+            rho_j[ j ] = np.nansum( value[ sel_inds ] * stratified_weights ) / np.nansum( stratified_weights )
```
leewujung, line 82: In what case would `stratified_weights` be NaN? It would have to come from `distance[ sel_inds ]`, so I guess if `distance` came from a diff operation, there could be NaN, but I have not tried to trace the code further up.

leewujung (approved):
@brandynlucca : Thanks for the changes. I realized that I did not review the function `stratified_transect_statistic`, so I just read through it now and only had a few minor suggestions and questions about NaN values.

The main higher-level question I have is on enabling `stratified_transect_statistic` to use 
both INPFC and KS stratum. It seems that the only place needing change (ie adding an if-else case) would be the resampling part that needs to choose from a list from that "stratification1" sheet (I forgot what you called it in the dataframes) rather than from this np.range array.

If there is not already an issue tracking this, we can add one and add this as one of the TODO items.

Otherwise I think this PR is ready to be merged once the small changes are made. :)

Also, parking some thoughts here regarding documentation: I think we can include the equations implemented here into the docs with references, so that we could stop the confusion or ambiguity once and for all. Let's discuss this more once the code is more settled.

EchoPro/computation/statistics.py
brandynlucca: When developing I think this was the case, but adjustments to `calculate_transect_distance` may deprecate the need for tip-toeing around NaN values here. I know that `numpy.nansum` adds computational costs -- so this may be worth fully evaluating later on. I'll add this as an issue.

PR #195, opened 2024-03-01
Add documentation for various parts of software

brandynlucca:
- Added folder for doc images
- Added placeholder markdown files for documentation discussing theoretical
information and mathematical equations
- Added example image to core_data_structure which was renamed
- `glossary.md` was added to contain a list of symbols and variable names
that can be found throughout the workflow of the software both programmatically and in the mathematical equations


leewujung:
This should be ready to go now!

One small suggestion I have is for the title "Mathematical implementation" -- it is weird to have the word "implementation" in this, since there is the algorithmic implementation. I suggest changing this to just "Theory" or "Underlying theory", since it includes both acoustics and geostatistics theories.

PR #198, opened 2024-03-06
Add unit tests + additional kriging updates 

brandynlucca:
This PR includes a variety of changes to the various kriging (and associated functions), as well as beginning to incorporate unit tests into the software.


EchoPro/tests/test_strata_mean_sigma_bs.py
```diff
@@ -0,0 +1,136 @@
...
+    assert all_length_df.shape[ 0 ] == 38
+
+    #----------------------------------
+    ### TS-length parameterization & modeling
+    # Fit parameters
+    ts_length_parameters = objS.config[ 'TS_length_regression_parameters' ][ 'pacific_hake' ]
+    slope = ts_length_parameters[ 'TS_L_slope' ]
+    intercept = ts_length_parameters[ 'TS_L_intercept' ]
+    assert slope == 20.0
+    assert intercept == -84.1
```
leewujung, line 83: Here you could remove these, since these values are set manually in this test and not from the codebase.
```suggestion
```

EchoPro/tests/test_strata_mean_sigma_bs.py
```diff
@@ -0,0 +1,136 @@
...
+
+    # Predict `TS`
+    all_length_df[ 'TS' ] = ts_length_regression( all_length_df[ 'length' ] , slope , intercept )
+    assert np.isclose( np.median( all_length_df[ 'TS' ] ) , -49.675 )
```
leewujung, line 87: If you want you could use what you have here to add a quick unit test `test_ts_length_regression`, since it needs only spec_df_reframed, length_df_copy, slope, intercept that you already have above.
`mock_survey` is not needed there, and since it's a unit test I would also suggest skipping the selection in `length_df_copy`.

EchoPro/tests/test_strata_mean_sigma_bs.py
```diff
@@ -0,0 +1,136 @@
...
+    ### Calculate mean `sigma_bs` per `haul_num` and then `stratum_num`
+    # `haul_num`
+    mean_haul_sigma_bs = (
+        all_length_df
+        .groupby( [ 'haul_num' , 'stratum_num' , 'species_id' ] )[ [ 'sigma_bs' , 'length_count' ] ]
+        .apply(lambda x: np.average( x[ 'sigma_bs' ] , weights= x[ 'length_count' ] ) )
+        .to_frame( 'sigma_bs_mean' )
+        .reset_index( )
+    )
+    assert mean_haul_sigma_bs.shape[ 0 ] == 14
```
leewujung, line 103: Here I would also suggest testing for the exact values, in addition to the shape.

leewujung:
We won't merge this and will just keep it for reference. Will close once tests are updated.

PR #206, opened 2024-03-08
docs: Add biomass conversion and kriging equations

brandynlucca:
Included are the first iterations of the RTD documentation for the equations and underlying theory for the calculations included within the `echopop` software.


leewujung (commented):
This PR needs more work, since the images are not cropped and rotated correctly and the sizes are not set. I will get that done this afternoon.

I will rename the PR since the biological apportioning components are not included. I will merge this and add an issue to track that as a high priority item.

PR #215, opened 2024-03-27
Update kriging apportionment method and functions

brandynlucca:
This includes: 

- Refactoring code to make the apportionment function easier to read
- Continue translating the original Matlab code to complete the total biomass calculation (which still requires refactoring and formatting, but the code is otherwise present)
- Begin adding unit tests for the kriging apportionment functions


leewujung:
Just a note: for this one is fine, but for the next ones please do it from your fork.

echopop/computation/biology.py
```diff
@@ -104,83 +128,200 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    """ 
+
+    ### Process the specimen data 
+    specimen_binned = (
         specimen_data
+        # ---- Drop unaged fish 
+        # !!! TODO: pending what FEAT says, weights associated with 
+        # missing ages should be added into the 'unaged' category. 
+        # This would further mean apportioning these into `weight_strata_aged_uanged`
         .dropna( how = 'any' )
```
leewujung, line 191: May be a good idea to specify the column to avoid accidental drops
```suggestion
        .dropna( how = 'any' , subset=AGE_COL)
```

echopop/computation/biology.py
```diff
@@ -104,83 +128,200 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        .loc[ lambda df: df.sex != 'unsexed' ]
+        # ---- Bin length
+        .bin_variable( length_intervals , 'length' )
+        # ---- Age bin
         .bin_variable( age_intervals , 'age' )
-        .bin_variable( length_intervals , 'length' )      
-        .groupby( [ 'stratum_num' , 'age_bin' , 'length_bin' , 'sex' ] )
-        .apply( lambda df: pd.Series( { 'summed_weight_all': df.weight.sum( ) ,
-                                        'summed_weight_adult': df.loc[ df.age > 1 ].weight.sum( ) } ) ) 
+    )
```
leewujung, line 198: The lambda function could just be a row selection, and if I were to write it often I just write the variables out explicitly:
```python
specimen_binned = specimen_data[specimen_data["sex"]=="unsexed"].dropna(how="any", subset="...")
specimen_binned["length"] = specimen_binned.bin_variable(length_intervals, "length")
specimen_binned["age"] = specimen_binned.bin_variable(age_intervals, "age")
```

leewujung: And in what you have for `bin_variable`:
I would vectorize it to do:
```python
dataframe[f"{bin_variable}_bin"] = pd.cut(x[bin_variable], bin_values)
```
instead of using lambda functions.

echopop/computation/biology.py
```diff
@@ -104,83 +128,200 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        specimen_binned
+        # ---- Group weight summations across distributions of length and age 
+        # ---- for each sex within each stratum
+        .groupby( [ 'stratum_num' , 'sex' , 'length_bin' , 'age_bin' ] )
+        # ---- Sum the weights 
+        .apply( lambda df: pd.Series( { 'weight_all': df.weight.sum( ) ,
+                                        'weight_adult': df.loc[ df.age > 1 ].weight.sum( ) } ) ) 
+        # ---- Fill empty/non-existent values with 0's
+        .fillna( 0 )
         .reset_index( )
```
leewujung, line 211: For this I would likely do something like:
```python
specimen_binned_weight["weight_all"] = specimen_binned.groupby( ["stratum_num" , "sex", "length_bin", "age_bin"] )["weight"].sum()
specimen_binned_weight["weight_adult"] = specimen_binned[specimen_binned["age"] > 1].groupby( [ 'stratum_num' , 'sex' , 'length_bin' , 'age_bin' ] )["weight"].sum()

```

echopop/survey.py
```diff
@@ -1264,98 +1265,235 @@ def apportion_kriged_biomass( self ,
...
+            .copy( )
+            .merge( distributed_aged_weight_proportions , on = [ 'stratum_num' ] )
+            # ---- Distribute biomass estimates for each sexed bin among aged fish
+            # !!! TODO: this current used `B_a_adult_mean` for both calculations
+            .assign( biomass_sexed_aged_all = lambda df: ( df.B_a_adult_mean * 
+                                                     df.cell_area_nmi2 * 
+                                                     df.normalized_proportion_weight_sex_all ) ,
+                     biomass_sexed_aged_adult = lambda df: ( df.B_a_adult_mean * 
+                                                       df.cell_area_nmi2 * 
+                                                       df.normalized_proportion_weight_sex_adult )
```
leewujung, line 1325: Similar to the above, you can vectorize this to be:
```python
df["biomass_sexed_aged_all"] = (
  df["B_a_adult_mean"]
  * df["cell_area_nmi2"]
  * df["normalized_proportion_weight_sex_all"]
)
df["biomass_sexed_aged_adult"] = (
  df["B_a_adult_mean"]
  * df["cell_area_nmi2"]
  * df["normalized_proportion_weight_sex_adult"]
)
```

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,159 @@
...
+    # ---- `object_weight_strata`
+    check_values = np.array( [ 56.663158 , 9.231579 , 93.700000 , 31.568421 , 31.636842 ] )
+    assert np.allclose( check_values , object_weight_strata.weight_stratum_all )
+
+    # ---- `object_weight_strata_aged_unaged`
+    check_values = np.array( [ 51.400000 , 0.600000 , 81.700000 , 16.200000 , 12.900000 ,
+                            5.263158 , 8.631579 , 12.000000 , 15.368421 , 18.736842  ] )
+    assert np.allclose( check_values , object_weight_strata_aged_unaged.stratum_weight )
+
+def test_compute_summed_aged_proportions( mock_survey ):
```
leewujung, line 64: It seems that you're writing test for `compute_index_aged_weight_proportions` instead?

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,159 @@
...
+
+    ### Check shape 
+    test_dimensions = (
+        len( np.unique( specimen_binned.age_bin ) ) *
+        len( np.unique( specimen_binned.length_bin ) ) *
+        len( np.unique( specimen_binned.sex ) ) *
+        len( np.unique( specimen_binned.stratum_num ) )
+    )
+
+    assert obj_props_wgt_len_age_sex.shape == tuple( [ test_dimensions , 10 ] )
```
leewujung, line 156: I like that you're checking dimensions!

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,159 @@
...
+
+    ### Check values
+    assert obj_props_wgt_len_age_sex.equals( proportions_weight_length_age_sex )
```
leewujung, line 159: I think it may make sense to simplify the input data a bit (e.g., reduce the number of strata to 2, reduce the number of bins to 3, etc) and actually hard-code the expected values for the output of each sections: with a reduced input dataframe, you can easily hand-calculate the values of `specimen_binned`, and it is also easy to calculate `specimen_binned_weight`, `proportions_weight_length_age_sex`, and the final output.

The reason is that here you are just repeating the code blocks you have in the function, so if you update the function, it is likely you would update these sections too, which result into the test not actually functioning.

leewujung (changes requested):
@brandynlucca : I did a half pass on the code and tests you have here. I commented on a few blocks where you could further vectorize the ops. I did not go through all code blocks in `survey.py`, but I think the examples would give you the idea of what I meant. 

For the test, I think it would be good to actually hard-code the expected values, and not repeating the code blocks, to make sure the test would catch any discrepancies. Often for tests you reduce the test dataset to be small or much simpler than realistic data, so that you can catch implementation bugs and edge cases.

echopop/computation/biology.py
brandynlucca: @leewujung I agree with vectorizing the `bin_variable` function (i.e. removing the `lambda` function component); however, there is an issue with implementing the first comment. One minor issue is that `bin_variable` is still associated with the monkey patch decorator: `@patch_method_to_DataFrame( pd.DataFrame )` ([#166](https://github.com/OSOceanAcoustics/echopop/issues/166)). So the expected output is, effectively, the updated dataframe with the newly assigned bin variables. If the monkey patch is removed, then this would require `pipe(...)`. The monkey patch aside, the current implement therefore precludes assigning the output of `specimen_bined.bin_variable( ... )` to a given column as exemplified in your comment. 

The original intent of the `bin_variable` function was to coincide with `bin_stats(...)` to be a bit more explicit in describing what was occurring at that point in the dataframe pipe. So maybe this is more of a style/preference thing? Functionally, `bin_variable` could be circumvented entirely here by: 

```python
specimen_binned = specimen_data[ specimen_data.sex !="unsexed" ].dropna( how = "any" , subset="age" ) 
specimen_binned[ "length_bin" ] = pd.cut( specimen_binned.length , length_intervals ) 
specimen_binned[ "age_bin" ] = pd.cut( specimen_binned.age , age_intervals ) 
```
So this may indicate that `bin_variable` could be deprecated (or removed completely). I actually think `bin_variable` could be refitted to bin over multiple variables to cut down on repetitiveness both here and elsewhere in the dataflow. I'll add a `TODO` to `bin_variable` to make a note that either way something should be changed or that the function can be deprecated and updates are required to other functions/operations that use `bin_variable`. Thoughts?

echopop/computation/biology.py
brandynlucca: @leewujung Ah, related to what we had briefly chatted about when we met on 27 March 2024 (for reference): I see the virtue in vectorizing the columns this way, although my eyes inexplicably prefer generating both columns simultaneously (i.e. via `apply` within the dataframe chain). In this particularly case, this doesn't work given how I have the dataframes designed. The dimensions for `specimen_binned` (3253, 13) doesn't match the output of `specimen_binned.groupby( ["stratum_num" , "sex", "length_bin", "age_bin"] )["weight"].sum().reset_index()` (15840, 5), which represents the sums of weights across $N_I$ x $N_\alpha$ x $N_\ell$ x $N_s$ (which yields 15840 unique combinations of all four indices). I think `specimen_binned` may, subequently, be a misleading name and should be amended because while it contains the columns with the particular `length_bin` and `age_bin` intervals, the weights themselves are not actually quantized/summed/etc. In a related commit I have tried to rename this accordingly. Otherwise, I think the setup toward this step could also be tweaked so the code isn't so reliant on pipes (and therefore easier to vectorize). Happy to discuss this further as well!

echopop/computation/biology.py
leewujung: Good question! It's often hard to determine if something ops should be factored out as a function or not. I think given the usecase of `bin_variable`, if I were to do it, I would likely go with the path of not adding a new function, because `pd.cut` is widely used, so the intention for the code is clear. Also because the code for using `pd.cut` directly short.

echopop/computation/biology.py
leewujung: I am not following your description, but once you have the tests I can evaluate the test function to see what you're referring to here.

brandynlucca:
@leewujung: I had to do some additional debugging of the test functions since some things got overhauled as a consequence of of our discussion during our meeting on Monday. But there are now tests implemented for all functions associated with the kriging apportionment introduced in this PR, so I believe this is ready for you to take another look at.

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,878 @@
...
+    ### Check shape 
+    assert obj_props_wgt_len_age_sex.shape == expected_dimensions
+
+    ### Check datatypes
+    assert np.all( obj_props_wgt_len_age_sex.dtypes == expected_output.dtypes )
+    
+    ### Check data value equality
+    assert expected_output.equals( obj_props_wgt_len_age_sex )
+    
+def test_compte_aged_unaged_proportions( ):
```
leewujung, line 150: just fixing a tiny typo
```suggestion
def test_compute_aged_unaged_proportions( ):
```

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,878 @@
...
+                            np.linspace( 9 , 21 , 3 ) ) ,
+        'age_bin': pd.cut( np.tile( [ 1 , 2 ] , 8 ) , 
+                        np.array( [ 0.5 , 1.5 , 2.5 ] ) ) ,
+        'weight_all': [ 3.0 , 0.0 , 0.0 , 7.0 , 2.0 , 0.0 , 0.0 , 8.0 ,
+                        4.0 , 0.0 , 0.0 , 6.0 , 1.0 , 0.0 , 0.0 , 9.0 ] ,
+        'weight_adult': [ 0.0 , 0.0 , 0.0 , 7.0 , 0.0 , 0.0 , 0.0 , 8.0 ,
+                        0.0 , 0.0 , 0.0 , 6.0 , 0.0 , 0.0 , 0.0 , 9.0 ] ,
+        'total_weight_sex_all': np.repeat( 10 , 16 ).astype( np.float64 ).flatten( ) ,
+        'total_weight_sex_adult': [ 7.0 , 7.0 , 7.0 , 7.0 , 8.0 , 8.0 , 8.0 , 8.0 ,
+                                    6.0 , 6.0 , 6.0 , 6.0 , 9.0 , 9.0 , 9.0 , 9.0 ] ,
```
leewujung, line 223: I am confused about what these quantities are -- maybe it's time to put together a table for variation/column names in different part of the data structure? 
I think I am confused because `weight_adult` is 0 for the second row, but `total_weight_sex_adult` has a non-zero value. So I felt this goes back to I do not know what these quantities are.

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,878 @@
...
+        'sex': np.tile( [ 'female' , 'male' ] , 2 ).astype( object ) ,
+        'proportion_aged_weight_all': [ 0.10 , 0.10 , 0.20 , 0.20 ] ,
+        'proportion_aged_weight_adult': [ 0.05 , 0.05 , 0.10 , 0.10 ] ,
+        'proportion_weight_all': [ 0.050 , 0.050 , 0.040 , 0.040 ] ,
+        'proportion_weight_adult': [ 0.035 , 0.040 , 0.024 , 0.036 ] 
+    } )
+
+    ### Evaluate for later comparison 
+    eval_distributed_aged_weight_proportions = distribute_aged_weight_proportions( test_proportions_weight_length_age_sex ,
+                                                                                   test_aged_sex_proportions )
```
leewujung, line 304: I am confused about what `proportions_weight_length_age_sex` and `aged_sex_proportions` are at the input of this function. I think this is because the proportions do not sum to 1 in the columns I expected them to, so I can't tell if `proportions_weight_length_age_sex` is the weight proportion of each length and age bin WITHIN each sex, or if it is the weight proportion of each sex, length, and age bin. I have similar question about `aged_sex_proportions`... 

Maybe it would be the best if there is a table that maps these quantities to what is in the documentation?

echopop/survey.py
```diff
@@ -1264,98 +1272,113 @@ def apportion_kriged_biomass( self ,
...
-            kriged_results
-            .merge( average_weight_df , on = [ 'stratum_num' ] )
-            .assign( N_adult_kriged = lambda df: df.B_adult_kriged / df.average_weight_total )
-            .merge( self.acoustics[ 'sigma_bs' ][ 'strata_mean' ] , on = [ 'stratum_num' ] )
-            .assign( NASC_kriged = lambda df: df.N_adult_kriged * df.sigma_bs_mean )
+        ### Sum weights for aged/unaged and all within each stratum
+        weight_strata_aged_unaged , weight_strata = sum_strata_weight( haul_spp_matched ,
+                                                                       specimen_spp )
+        
+        ### Calculate the summed aged proportions for all aged fish, and for aged fish
```
leewujung, line 1322: do you mean for aged and UNaged fish?
```suggestion
        ### Calculate the summed aged proportions for all aged fish and unaged fish        
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+
+    Parameters
+    ----------
+    specimen_data: pd.DataFrame
+        Dataframe containing specimen weights
+    weight_strata: pd.DataFrame
+        Dataframe contained summed weights of both aged and unaged fish
+    """   
+    ### Calculate adult:all age ratio for each stratum
+    # ---- Drop unaged fish
```
leewujung, line 181: thought might be a good idea to be specific from which station of data these unaged fish are dropped
```suggestion
    # ---- Drop unaged fish from station 1 data (where most fish are aged)
```

echopop/computation/operations.py
```diff
@@ -4,7 +4,18 @@
...
+# ==== Currently works with dataframes as a whole, but 
+# ==== it is probably more efficient/sensible to refit 
+# ==== this function to create an array of bin intervals that 
+# ==== can then be used to define a dataframe column. It may 
+# ==== also make more sense to simply remove this function entirely
+# ==== since it's not particularly onerous to directly use 
+# ==== `pd.cut(...)`. If this function were to stay, it'd probably 
+# ==== be usefule to enable multiple `bin_values` and `bin_variable`
+# ==== inputs so that the function doesn't need to be repeated for 
+# ==== multiple bin variables like age and length
```
leewujung, line 18: maybe put these in an issue so it's not buried in code?

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        1.0 - aged_proportions.proportion_aged_weight_adult
+    )
+
+    return aged_proportions.drop( [ 'weight_ratio' ] , axis = 1 )
+
+def compute_aged_weight_proportions( specimen_data: pd.DataFrame ,
+                                     length_intervals: np.ndarray ,
+                                     age_intervals: np.ndarray ):
+    """
+    Calculate length-age binned weight proportions
```
leewujung, line 224: Thought it may be a good idea to include the data source
```suggestion
    Calculate length-age binned weight proportions based on specimen_df
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    age_intervals: np.ndarray
+        Array containing age bins/intervals
+    """ 
+
+    ### Process the specimen data 
+    # ---- Drop unaged fish
+    # ==== !!! TODO: pending what FEAT says, weights associated with 
+    # ==== missing ages should be added into the 'unaged' category. 
+    # ==== This would further mean apportioning these into `weight_strata_aged_uanged`
+    specimen_data_filtered = specimen_data[ specimen_data.sex != 'unsexed' ].dropna( how = 'any' , subset = 'age' )
```
leewujung, line 241: This does 2 things: select only sexed fish and dropped unaged fish, right?

Are there other place where unsexed fish from station 1 is used? If not, maybe we could also drop unsexed fish from station 1 when `specimen_df` is first read in. Or is it that there are too many unsexed fish so it is not a good idea to just drop them at data loader?

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    age_distribution: np.ndarray
+        Array containing age bins/intervals
+    aged_proportions: pd.DataFrame
+        Dataframe containing the weight proportions of aged fish within each stratum
+    """
+
+    ### Sum aged fish weights across age and length bins, then calculate weight proportions
+    proportions_weight_length_age_sex = compute_aged_weight_proportions( specimen_data ,
+                                                                         length_distribution ,
+                                                                         age_distribution )
```
leewujung, line 402: Add to the comment to make it specific that it is for age and length weight proportion _within_ each sex. Otherwise the output variable name does not make sense.

You might want to consider renaming `proportions_weight_length_age_sex` to something like `proportions_weight_length_age_within_each_sex` so that there is no chance of confusion.
```suggestion
    ### Sum aged fish weights across age and length bins for each sex, then calculate weight proportions within each sex
    proportions_weight_length_age_sex = compute_aged_weight_proportions( specimen_data ,
                                                                         length_distribution ,
                                                                         age_distribution )
```

echopop/tests/test_kriging_biomass_apportionment.py
```diff
@@ -0,0 +1,878 @@
...
+    #----------------------------------
+    ### `eval_aged_sex_proportions`
+    # ---- Shape
+    assert eval_aged_unaged_weight_proportions.shape == expected_dimensions
+    # ---- Datatypes
+    assert np.all( eval_aged_unaged_weight_proportions.dtypes == expected_output.dtypes )
+    # ---- Dataframe equality
+    assert np.allclose( eval_aged_unaged_weight_proportions , expected_output )
+
+def test_compute_aged_sex_proportions( ):
```
leewujung, line 205: Making it identical to the function name
```suggestion
def test_compute_aged_sex_weight_proportions( ):
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
-def normalize_length_age_sex_weight_proportions( weight_length_age_sex_stratum: pd.DataFrame ,
-                                                 weight_strata: pd.DataFrame ):
+    ### Return output
+    return proportions_weight_length_age_sex
+
+def compute_aged_sex_weight_proportions( proportions_weight_length_age_sex: pd.DataFrame ,
+                                         aged_proportions: pd.DataFrame ):
+    """
+    Compute the aged proportions across all fish and specific sexes for each
+    length and age bin
```
leewujung, line 297: This function doesn't actually get into the length and age bin, that is in the previous function I think.
```suggestion
def compute_aged_sex_weight_proportions( proportions_weight_length_age_sex: pd.DataFrame ,
                                         aged_proportions: pd.DataFrame ):
    """
    Compute the weight proportion for aged/unaged fish of either sex (2x2 choice) across all fish.
    Also compute the weight proportion of only adult aged/unaged fish of either sex across all fish.
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
-        .reset_index( ) 
-    ) 
+    # ---- Normalized weight proportions for all fish
+    distributed_aged_weight_proportions[ 'normalized_proportion_weight_all' ] = (
+        distributed_aged_weight_proportions.proportion_weight_sex_all * distributed_aged_weight_proportions.proportion_weight_all
+    )
 
-    ### Carriage return
-    return dist_weight_sum
+    # ---- Normalized weight proportions for jus adult fish
```
leewujung, line 368:
```suggestion
    # ---- Normalized weight proportions for just adult fish
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    
+    ### Calculate the relative weight proportions of each length-age bin for each sex within each stratum
+    proportions_weight_length_age_sex = (
+        specimen_binned_weight
+        # ---- Calculate total sex-specific weights for each stratum
+        .assign( total_weight_sex_all = lambda df: df.groupby( [ 'stratum_num' , 'species_id' , 'sex'  ] )[ 'weight_all' ].transform( sum ) ,
+                 total_weight_sex_adult = lambda df: df.groupby( [ 'stratum_num' , 'species_id' , 'sex' ] )[ 'weight_adult' ].transform( sum ) )
+        # ---- Calculate the weight proportions within each sex: from Matlab --> Len_age_key_wgt_*n
+        .assign( proportion_weight_sex_all = lambda df: df.weight_all / df.total_weight_sex_all ,
+                 proportion_weight_sex_adult = lambda df: df.weight_adult / df.total_weight_sex_adult )
```
leewujung, line 278: It'll probably be a good idea to change `weight_all` to something like `weight_sex_length_age_all` so that it is clear that these are grouped separately, for readability.

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
-        .groupby( [ 'stratum_num' , 'sex' ] ) 
-        .apply( lambda df: pd.Series( {  
-            'proportion_normalized': ( df.proportion_weight_all * ( df.summed_weight_all / df.weight_stratum_total ).sum( ) ).sum( ) 
-        } ) ) 
-        .reset_index( ) 
-    ) 
+    # ---- Normalized weight proportions for all fish
+    distributed_aged_weight_proportions[ 'normalized_proportion_weight_all' ] = (
+        distributed_aged_weight_proportions.proportion_weight_sex_all * distributed_aged_weight_proportions.proportion_weight_all
+    )
```
leewujung, line 366: Here it would be good to write out what `proportion_weight_sex_all` and `proportion_weight_all` contains, because the variable names themselves do not contain that info.

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    ### Calculate the summed aged proportions for all aged fish, and for aged fish 
+    # ---- belonging to each sex
+    aged_sex_proportions = compute_aged_sex_weight_proportions( proportions_weight_length_age_sex ,
+                                                                aged_proportions )
+    
+    ### Calculate weight proportions of aged fish distributed over age-length bins
+    # ---- for each sex within each stratum relative to the summed weights of each
+    # ---- stratum (i.e. aged + unaged weights)
+    distributed_sex_proportions = distribute_aged_weight_proportions( proportions_weight_length_age_sex ,
+                                                                      aged_sex_proportions )
```
leewujung, line 413: may be a good idea to make `distributed_sex_proportions` something longer, like `distributed_sex_length_age_proportions_wrt_all`

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
-                                weight_strata: pd.DataFrame ,
-                                weight_strata_station: pd.DataFrame ,
-                                aged_proportions: pd.DataFrame ,
-                                length_intervals: np.ndarray ):
+    ### Apportion sexed biomass across age-length
+    # ---- Initialize dataframe
+    stratum_sexed_kriged_biomass = distributed_sex_proportions.set_index( 'stratum_num' )
+    stratum_sexed_kriged_biomass[ 'biomass_kriged' ] = kriged_stratum_biomass
+
+    # ---- Apportioned biomass for all fish
```
leewujung, line 423:
```suggestion
    # ---- Apportioned biomass for all fish to different sex-length-age bins
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        * stratum_sexed_kriged_biomass.normalized_proportion_weight_all
+    ).fillna( 0 )
+
+    # ---- Apportioned biomass for adult fish
+    stratum_sexed_kriged_biomass[ 'biomass_sexed_aged_adult' ] = (  
+        stratum_sexed_kriged_biomass.biomass_kriged
+        * stratum_sexed_kriged_biomass.normalized_proportion_weight_adult
+    ).fillna( 0 )
+
+    # ---- Sum across strata to produce 'grand totals' for each sex
```
leewujung, line 435: This seems not accurate, since underneath the length and age bins are separate?

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    ).fillna( 0 )
+
+    # ---- Sum across strata to produce 'grand totals' for all fish
+    apportioned_total_kriged_biomass = (
+        stratum_kriged_biomass
+        .groupby( [ 'species_id' , 'length_bin' , 'age_bin' ] )
+        .agg( { 'biomass_aged_all': 'sum' ,
+                'biomass_aged_adult': 'sum' } )
+        .reset_index( )
+    )
```
leewujung, line 506: I am not sure why you need to go from L446 all the way here to calculate these quantities. Aren't these just sums over both sexes for each length-age bin, that you can compute directly from `apportioned_sexed_kriged_biomass`?

If there are special cases that prevents going from `apportioned_sexed_kriged_biomass` to `apportioned_total_kriged_biomass`, the test should reflect that.

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        .agg( { 'weight_all': 'sum' ,
+                'weight_adult': 'sum' } )
+        .reset_index( )
+    )
+
+    # ---- Sum stratum weights
+    # ---- All fish
+    proportions_weight_length_age[ 'total_weight_all' ] = (
+        proportions_weight_length_age.groupby( [ 'stratum_num' , 'species_id' ] )[ 'weight_all' ].transform( sum )
+    )
```
leewujung, line 458: I got the following warning for calls like this one:
```
echopop/echopop/computation/biology.py:457: FutureWarning: The provided callable <built-in function sum> is currently using SeriesGroupBy.sum. In a future version of pandas, the provided callable will be used directly. To keep current behavior pass the string "sum" instead.
    proportions_weight_length_age.groupby( [ 'stratum_num' , 'species_id' ] )[ 'weight_all' ].transform( sum )
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        .reset_index( )
+    )
+
+    ### Return output (tuple)
+    return apportioned_sexed_kriged_biomass , apportioned_total_kriged_biomass
+
+def compute_unaged_number_proportions( length_data: pd.DataFrame ,
+                                       length_intervals: np.ndarray ):
+    """
+    Calculate length binned weight proportions among unaged fish
```
leewujung, line 514:
```suggestion
    Calculate length binned number proportions among unaged fish  
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+        Dataframe containing the modeled weights for each sex and length bin
+        fit from the length-weight regression (fit for all animals)
+    aged_proportions: pd.DataFrame
+        Dataframe containing the weight proportions of unaged fish within each stratum
+    """
+
+    ### Calculate number proportion
+    proportions_unaged_length = compute_unaged_number_proportions( length_data , length_distribution )
+
+    ### Calculate the normalized weight per unit length distribution (W_Ln_ALL in the original Matlab code)
```
leewujung, line 720: I would suggest changing this to avoid confusion because I don't think Chu's description was accurate:
```suggestion
    ### Calculate the weight proportion of unaged fish of each length bin (W_Ln_ALL in the original Matlab code)
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    ### Sum weight proportions for each strata to get the weight per length bin
+    # ---- Calculate the summed weights per stratum (`w_ln_array_sum` in the original Matlab code)
+    proportions_unaged_weight_length[ 'weight_stratum_proportion' ] = (
+       proportions_unaged_weight_length.groupby( [ 'stratum_num' ] )[ 'weight_proportion_all' ].transform( sum )
     )
+    
+    # ---- Calculate the weight per length distribution bin (`w_ln_all_N` in the original Matlab code)
+    proportions_unaged_weight_length[ 'proportion_weight_length' ] = (
+        proportions_unaged_weight_length.weight_proportion_all / proportions_unaged_weight_length.weight_stratum_proportion
+    )
```
leewujung, line 598: Some suggestions to avoid confusion of the quantities:
```suggestion
    # ---- Calculate the estimated weight for each length bin within each stratum (`w_ln_all_array` in the original Matlab code)
    proportions_unaged_weight_length[ 'weight_proportion_all' ] = (
        proportions_unaged_weight_length.weight_modeled * proportions_unaged_weight_length.proportion_number_all
    )

    ### Calculate weight proportions for each length bin for all strata
    # ---- Calculate the summed estimated weights across all length bins within each stratum (`w_ln_array_sum` in the original Matlab code)
    proportions_unaged_weight_length[ 'weight_stratum_proportion' ] = (
       proportions_unaged_weight_length.groupby( [ 'stratum_num' ] )[ 'weight_proportion_all' ].transform( sum )
    )
    
    # ---- Calculate the weight proportion of each length bin within each stratum (`w_ln_all_N` in the original Matlab code)
    proportions_unaged_weight_length[ 'proportion_weight_length' ] = (
        proportions_unaged_weight_length.weight_proportion_all / proportions_unaged_weight_length.weight_stratum_proportion
    )
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    ).fillna( 0 )
+
+    ### Filter out unnecessary columns and return output
+    return proportions_unaged_length.filter( regex = '^(?!number_|stratum_number_).*' )
+
+def compute_unaged_weight_proportions( proportions_unaged_length: pd.DataFrame ,
+                                       length_weight_df: pd.DataFrame ):
+    """
+   Compute the unaged weight proportions across all fish and specific sexes for each
+   length bin
```
leewujung, line 562: need to fix indent -- let's use the code formatter!

leewujung, line 562: wording suggestions and also fixed indent -- let's use the code formatter!
```suggestion
    Compute the weight proportion of each length bin within each stratum for all unaged fish.
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
-            .reset_index( drop = True ) ) ) 
-        .merge( aged_proportions.loc[ : , [ 'stratum_num' , 'proportion_unaged_total' ] ] ,  
-                on = [ 'stratum_num' ] )         
-        .assign( proportion_unaged_sex = lambda x: x.sex_proportion * x.proportion_unaged_total ) 
-    ) 
+        .apply( lambda df: pd.Series( {
+                'weight_interp': ( np.interp( df.length ,
+                                              df.length_bin_value ,
+                                              df.weight_modeled ) * df.length_count ).sum( )
+            } ) )
```
leewujung, line 652: Let's not use interpolation but use the regression relationship to compute the estimated weight?

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+
+    ### Drop unnecessary columns and return output
+    return proportions_unaged_weight_length.filter( regex = '^(?!weight_|proportion_number_).*' )
+
+def compute_unaged_sex_proportions( length_data: pd.DataFrame ,
+                                    length_intervals: pd.DataFrame ,
+                                    length_weight_df: pd.DataFrame ,
+                                    aged_proportions: pd.DataFrame ):
+    """
+    Calculate the normalized weight proportions for unaged sexed fish
```
leewujung, line 608: Trying to get rid of the "normalized" wording since proportions are normalized already.
```suggestion
    Calculate the weight proportion of different sexes within the unaged fish samples.
```

echopop/computation/biology.py
```diff
@@ -104,160 +128,751 @@ def sum_strata_weight( haul_data: pd.DataFrame ,
...
+    )
+
+    # ---- ' Grand total ' for all sexed unaged fish
+    apportioned_kriged_biomass = (
+            stratum_kriged_biomass
+            .groupby( [ 'species_id' , 'length_bin' ] )
+            .agg( { 'biomass_unaged_all': 'sum' ,
+                    'biomass_unaged_adult': 'sum' } )
+            .reset_index( )
+        )
```
leewujung, line 811: Similar to my question in `compute_aged_biomass` [here](https://github.com/OSOceanAcoustics/echopop/pull/215/files?diff=unified&w=0#r1561905890):
wouldn't you be able to compute what's in `apportioned_kriged_biomass` by summing over all sexes in `apportioned_sexed_kriged_biomass`?

leewujung (changes requested):
Hey @brandynlucca : Thanks for the PR! I made various inline comments. Below are some overall comments:

Under the function `apportion_kriged_biomass`, the outputs `weight_strata_aged_unaged` and `unaged_biomass` appear unused  after they are created. Since you went all the way to the end for the apportioning, this means that these variables can be removed, right?

For dropping unaged fish in station 1, how about dropping those fish when you read in `specimen_df` in the data loader? This way we don't have to worry about dropping those separately in each functions. It would be as if those fish samples never exist.

I guess the above may not be true for unsexed fish, but in `compute_aged_weight_proportions` the unsexed fish are also dropped (not selected, [here](https://github.com/OSOceanAcoustics/echopop/pull/215/files?diff=unified&w=0#r1561878197)). Should these be removed at the data loader too? Or because the number is larger and these samples are used in some other processing assuming the unsexed fish are assumed to follow the same male:female proportion, we should not drop them at data loader?

Lastly, there are two types of function names `compute_*` and `calculate_*` under `biology.py` -- is that intentional? Seems that `compute_*` are functions that are called by `calculate_*`?

echopop/computation/biology.py
brandynlucca: - Drop unaged fish in data loader
- Adding caveat doc page related to the treatment of unsexed and removal of unaged fish
 - Which functions do the unsexed fish become relevant (e.g. length-weight regression)
- Print out in console in data loader: "x unaged samples removed from Station 2 dataset" or something to that effect

Issue #202, opened 2024-03-07
Resolve discrepancies in kriged areal biomass density calculations


brandynlucca:
Areal biomass density estimates produced from the kriging analysis tend to be substantially smaller than values produced in previous implementations of this software/analysis. This discrepancy further propagates into subsequently large coefficient of variation (_CV_) estimates. This is a priority issue that is being tracked and requires further dissection of the code and accompanying integration tests since the precise source of the issue has not yet been identified. 

- [ ] Incorporate tests for support functions associated with the kriging analysis
- [ ] Incorporate tests (on mocked data) for the kriging analysis
- [ ] Identify the source of the areal biomass density calculation discrepancy and patch the bug/error

This also relates to #187 and will apply to #185 for once the Jolly-Hampton (1990) algorithm is implemented for the kriging analysis.

leewujung:
This is being addressed in #215.

brandynlucca:
The difference in integrated transect biomass between `echopop` and `EchoPro` is < 0.10 kmt, which is a < 0.001% difference. This differentiation can be attributed to a slightly different weight distribution (across age and length) due to `EchoPro` having a maximum age that doesn't capture all fish and a bug in how the female-specific length-weight regression is calculated (which also incorporates unsexed fish into the calculation). These differences propagate into the apportioned kriged biomass estimates. These issues have been addressed in the `kriging_analysis(...)` `Survey`-class method reflected in PR #241.

This can be closed once PR #241 is merged.

PR #221, opened 2024-04-06
Add unit tests for `init`, `transect_analysis`, and `stratified_summary` Survey-class methods

brandynlucca:
This PR includes new unit tests for the following `Survey` class methods and related functions:

**`init`**
- `load_survey_data`
 - This does _not_ include a standalone test for the `validate_data_columns` and `read_validated_data` functions nested within this method 
- `biometric_distributions`

**`transect_analysis`**
- `fit_binned_length_weight_relationship`
- `strata_sex_weight_proportions`
- `strata_age_binned_weight_proportions`
- `nasc_to_biomass_conversion`
- Functions nested within each of these methods have also had new tests written for more specific/granular examples

**`stratified_summary`**
- Functions nested within each of these methods have also had new tests written for more specific/granular examples

**Other**
- Some functions such as those contained within `echopop.computation.operations` are used across multiple methods 
- Tests have been added for each of these (e.g. `group_merge`, `meld`, `stretch`) using both the current monkey-patched implementation _as well as_ the future updated use as 'normal' functions (i.e. `dataframe.method( args ) --> method( dataframe , args )`
- Additional utility test functions (i.e. `dictionary_shape_equal`) were added for testing equality in structures between nested dictionaries 
- Small tweaks were included to the actual functions to account for some tested edge cases that caused tests to fail, such as unexpected `NaN` outputs


brandynlucca:
@leewujung: okay, so after futzing with the library versions, I believe I have squared away the issues I was running into regarding the test failures online (but passing locally).

echopop/survey.py
```diff
@@ -1250,7 +1252,7 @@ def krige( self ,
         ### TODO: This should be refactored out as an external function 
         ###### rather than a Survey-class method.
         ### Apportion biomass based on age and sex
-        self.apportion_kriged_biomass( species_id )
+        # self.apportion_kriged_biomass( species_id )
```
leewujung, line 1255: I guess this will be something that gets resolved once #215 is merged?

echopop/tests/conftest.py
```diff
@@ -1,7 +1,13 @@
 import pytest
 from pathlib import Path
 from echopop import Survey
+from _pytest.assertion.util import assertrepr_compare
 
+def pytest_assertrepr_compare( config , op , left , right ):
```
leewujung, line 6: Is this a utility function that you call while debugging, or is this intended to be run every time? As of now it does not gets run every time since this function is not called.

echopop/tests/test_transect_functions.py
```diff
@@ -0,0 +1,311 @@
...
+
+    #----------------------------------
+    ### Run tests: `index_transect_age_sex_proportions`
+    #----------------------------------
+    ### Check shape 
+    assert eval_nasc_fraction_total_df.shape == expected_dimensions
+    ### Check datatypes
+    assert np.all( eval_nasc_fraction_total_df.dtypes == expected_output.dtypes )
+    ### Dataframe equality
+    assert np.allclose( eval_nasc_fraction_total_df , expected_output , rtol = 1e-1 )
```
leewujung, line 163: So the function I was telling you about was [`np.issubdtype`](https://numpy.org/doc/stable/reference/generated/numpy.issubdtype.html#numpy.issubdtype):
```python
# below will both pass
assert np.issubdtype(np.int32, np.integer)
assert np.issubdtype(np.int64, np.integer)
```

For the tests here, I think you can specify separate dicts for the expected dtype and expected values, like how you separate out the expected dimensions, and test the dtype and values separately:
```python
expected_dtype = {
    'longitude': np.floating,
    'transect_num': np.integer,
    'stratum_num': np.integer,
     ...
}
expected_values = {
    'longitude': np.repeat( [ -180.0 , -120.0 , -170.0 , -110.0 ] , 2 ) ,
    'transect_num': np.repeat( [ 1 , 2 , 3 , 4 ] , 2 ),
    'stratum_num': np.repeat( [ 0 , 1 ] , 4 ),
    ...
}

# test each column separately
for col in eval_nasc_fraction_total_df.columns:
    assert np.issubdtype(eval_nasc_fraction_total_df[col], expected_dtype[col])
    assert np.allclose(eval_nasc_fraction_total_df[col].data, expected_values[col])

```

You may need to use `np.allclose` to test for equality for `np.floating` types. And just a small note on what we discussed that rtol=1e-1 is a pretty big tolerance, and may signal that something may be going on underneath.

echopop/tests/test_data_transect_analysis.py
```diff
@@ -0,0 +1,633 @@
...
+def test_fit_binned_length_weight_relationship( mock_survey ):
+    
+    #### Pull in mock Survey object
+    objS = mock_survey
+    
+    ### Initialize objS for `length_weight`
+    objS.statistics[ 'length_weight' ] = { }
+    
+    ### Re-parameterize `specimen_df` with dummy data 
+    objS.biology[ 'specimen_df' ] = pd.DataFrame(
```
leewujung, line 14: You can specify the type (you'll see the methods and attributes turn yellow, which i think means that VSCode recognize these attributes/methods), and also use mock_survey directly in the test:
```suggestion
def test_fit_binned_length_weight_relationship( mock_survey: Survey ):
    
    ### Initialize objS for `length_weight`
    mock_survey.statistics[ 'length_weight' ] = { }
    
    ### Re-parameterize `specimen_df` with dummy data 
    mock_survey.biology[ 'specimen_df' ] = pd.DataFrame(
```

echopop/tests/test_operations.py
```diff
@@ -0,0 +1,450 @@
...
+    ### Run tests: `count_variable`
+    #----------------------------------
+    ### Check shape
+    assert eval_dataframe_monkey.shape == expected_dimensions
+    assert eval_dataframe_function.shape == expected_dimensions
+    ### Check output
+    assert np.all( eval_dataframe_monkey == expected_output )
+    assert np.all( eval_dataframe_function == expected_output )
+
+def test_group_merge( ):
```
leewujung, line 329: This comment does not really belong to this PR, but since this function shows up here...

It seems that `group_merge` is only used once, and the content is very specific. I think in this case the operations can be exposed in the calling function `index_transect_age_sex_proportions` so that the code is easy to follow.

leewujung (changes requested):
Hey @brandynlucca : The tests look great! I have only a couple embedded comments, and those below are applicable more generally so I put them in the overall comment here:

There are a few tests you could use the same approach for testing dtype and values separately like in [this comment](https://github.com/OSOceanAcoustics/echopop/pull/221/files#r1564869561).


Just a note on what we discussed that rtol=1e-1 is a pretty big tolerance, which may signal that something is going on underneath that is not quite "equal", since the [determination is based on the following](https://numpy.org/doc/stable/reference/generated/numpy.isclose.html):
> For finite values, isclose uses the following equation to test whether two floating point values are equivalent.
>
> absolute(a - b) <= (atol + rtol * absolute(b))

There are a few tests in which similar large rtol is used, like in `test_stratified_transect_statistic`, so it'd be good to check what's going on.


Lastly, I think you could move the functions in `utility_testing_functions.py` to `conftest.py` as fixtures, so that they can be reused in the tests -- see [Fixtures are reusable](https://docs.pytest.org/en/6.2.x/fixture.html#fixtures-are-reusable).

echopop/survey.py
brandynlucca: Yep! The plan is to uncomment this once the kriging rework is completed re: [#202](https://github.com/OSOceanAcoustics/echopop/issues/202).

PR #222, opened 2024-04-08
Updates and unit tests for grid coordinate transformation

brandynlucca:
This PR includes:

- New unit tests for `transform_geometry` and `standardize_coordinates`
- Minor refactoring and vectorization for the above

### DIFF
--- a/echopop/computation/spatial.py
+++ b/echopop/computation/spatial.py
@@ -183,20 +183,20 @@ def transform_geometry( dataframe: pd.DataFrame ,
     ### based on 'd_longitude' and 'd_latitude' inputs
     # ---- If 'd_longitude' and 'd_latitude' are not defined
     if ( d_longitude is None ) & ( d_latitude is None ):
-        d_longitude = geodataframe_geometry.geometry.x.max( ) - geodataframe_geometry.geometry.x.min( )
-        d_latitude = geodataframe_geometry.geometry.y.max( ) - geodataframe_geometry.geometry.y.min( )
+        d_longitude = geodataframe_geometry.geometry_transformed.x.max( ) - geodataframe_geometry.geometry_transformed.x.min( )
+        d_latitude = geodataframe_geometry.geometry_transformed.y.max( ) - geodataframe_geometry.geometry_transformed.y.min( )
 
     ### Standardize x- and y-coordinates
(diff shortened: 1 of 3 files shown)


echopop/computation/spatial.py
```diff
@@ -103,10 +103,8 @@ def calculate_transect_distance( dataframe ,
     )
     
 def transform_geometry( dataframe: pd.DataFrame ,
```
leewujung, line 105: Here the docstring is missing `range_out`.

After reading through the code though, I feel `range_out` should not be an input argument. I feel the code would be more readable if you always return a 3-element list, but in the range_out=False case, simply ignore the output. Like in `standardize_coordinates`, do:
```python
mesh_trans_df, _, _ = transform_geometry( ... )
```
And the other case the code does not need to change.

This will cut down the test case also.

leewujung (commented):
Hey @brandynlucca : The test looks good! I did a PR to your branch on the use of pytest.parametrized for this type of things. But then I also think you could remove `range_out` as an argument altogether.

Reading more, I think that `standardize_coordinates` may be a bit problematic. It does not make sense to me that the argument `adjust_mesh` contains a hidden selection that it updates `self.statistics[ 'kriging' ][ 'mesh_df' ]` only when `adjust_mesh=True`.

I couldn't find a function that calls the current `standardize_coordinates` so don't know what the use case is, but I feel `transform_geometry` should be called directly without another layer of hidden function, unless the same operation is done many many times. What I am thinking is that in the case when only  `self.biology[ 'population' ][ 'biomass' ][ 'biomass_age_df' ]` needs to be updated, do
```python
`self.biology[ 'population' ][ 'biomass' ][ 'biomass_age_df' ]` = transform_geometry(...)
```
and in the case when both `self.biology[ 'population' ][ 'biomass' ][ 'biomass_age_df' ]` and `self.statistics[ 'kriging' ][ 'mesh_df' ]` need to be updated, do:
```python
self.biology[ 'population' ][ 'biomass' ][ 'biomass_age_df' ], d_x, d_y = transform_geometry(...)
self.statistics[ 'kriging' ][ 'mesh_df' ] = transform_geometry(..., d_x, d_y)
```
This way there is no hidden operation.

PR #241, opened 2024-05-23
Overhaul to kriging methods, data structure, and primary `Survey`-class methods

brandynlucca:
This includes a variety of updates, bug-fixes, and new features. There are a few more things I need to clean up (documentation, additional in-line comments, etc) but that isn't explicitly related to the actual code functioning.


Data
```diff
@@ -1 +1 @@
-C:/Users/Brandyn/Documents/GitHub/EchoPro_data/2019_consolidated_files
\ No newline at end of file
+C:/Users/Brandyn/Documents/GitHub/EchoPro_data/2019_consolidated_files
```
leewujung, line 1: Probably a good idea to not push the symbolic link `Data`. If this is used somewhere (though it really shouldn't be in the long run), we can add how to set it up to the README.

echopop/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .survey import Survey
+from .utils import operations
 
-__all__ = ["Survey"]
+__all__ = ["Survey", "operations"]
```
leewujung, line 4: why do you need to expose `operations` at the root level when importing the package? It is not currently used in the example notebook.

echopop/statistics.py
```diff
@@ -0,0 +1,268 @@
...
+        "total": {
+            "estimate": np.mean(replicate_total),
+            "confidence_interval": confidence_interval(replicate_total),
+        },
+    }
+    # ---- Return outputs
+    return resampled_distributions, stratified_results
+
+
+def array_math(left_array: np.ndarray, right_array: np.ndarray, operation: str) -> np.ndarray:
```
leewujung, line 233: I would suggest removing this function and just write out the transpose needed in the code. One-off utility functions like this makes code harder to read and are prone to error, because you are essentially trying to overload binary operators in a somewhat limited way. This is also only used twice and the underlying operations are very simple.

echopop/acoustics.py
```diff
@@ -0,0 +1,452 @@
...
+
+            sigma_bs_stratum_impute.loc[
+                sigma_bs_stratum_impute[stratum_col] == i, "sigma_bs_mean"
+            ] = sigma_bs_indexed["sigma_bs_mean"].mean()
+
+    # Return output
+    return sigma_bs_stratum_impute
+
+
+def summarize_sigma_bs(
```
leewujung, line 121: How about renaming this to `calc_mean_sigma_bs`? Summarize does not convey the calculation of mean values.

echopop/acoustics.py
```diff
@@ -0,0 +1,452 @@
...
+    length_data: pd.DataFrame,
+    specimen_data: pd.DataFrame,
+    strata_dict: dict,
+    configuration_dict: dict,
+    settings_dict: dict,
+) -> dict:
+    """
+    Calculates the stratified mean sigma_bs for each stratum
```
leewujung, line 129:
```suggestion
    Calculates the stratified mean sigma_bs for each haul and each stratum
```

echopop/acoustics.py
leewujung: Also, I feel this function would be a lot simpler if you just handle one species only. I am not see error checking if the species in the TS-length relationship is different from what's in the specimen or the reverse, though I also did not try to dig in to see what the edge cases may be.

echopop/acoustics.py
```diff
@@ -0,0 +1,452 @@
...
+    # ---- Merge NASC measurements with mean sigma_bs for each stratum
+    nasc_biology = nasc_interval_df.merge(sigma_bs_strata, on=[stratum_col])
+    # ---- Calculate the number densities
+    nasc_biology["number_density"] = (
+        nasc_biology["fraction_hake"]
+        * nasc_biology["nasc"]
+        / (4.0 * np.pi * nasc_biology["sigma_bs_mean"])
+    )
+    # ---- Round to the nearest whole number (we don't want fractions of a fish)
+    nasc_biology["number_density"] = np.round(nasc_biology["number_density"])
```
leewujung, line 321: I think fraction is fine when we talk about density? Even if you force number density to be integer per nmi^2, it will still be non-integer per km^2, for example.

echopop/acoustics.py
```diff
@@ -0,0 +1,452 @@
...
+    # ---- Extract the sex-specific number proportions for each stratum
+    sex_stratum_proportions = analysis_dict["biology"]["proportions"]["number"][
+        "sex_proportions_df"
+    ].copy()
+    # ---- Filter out sexes besides 'male' and 'female'
+    sex_stratum_proportions = sex_stratum_proportions[
+        sex_stratum_proportions.sex.isin(["male", "female"])
+    ]
+    # ---- Merge with the NASC measurements
+    nasc_biology_sex = nasc_biology.merge(sex_stratum_proportions, on=[stratum_col, "species_id"])
```
leewujung, line 336: I may be missing something but I don't think it makes sense to separate out different species here. In the above you already have `hake_fraction`, which is hard coded to a single species. I have the same suggestion as in the function that calculates mean sigma_bs that it is simpler to just handle one species.

echopop/acoustics.py
```diff
@@ -0,0 +1,452 @@
...
+            configuration_dict["TS_length_regression_parameters"]["pacific_hake"],
+            settings_dict,
+        )
+        # ---- Calculate adult proportions
+        # -------- Initialize dataframe
+        adult_proportions = age1_proportions.copy()
+        # -------- Invert proportions
+        adult_proportions.iloc[:, np.arange(1, 4)] = (
+            1.0 - adult_proportions.iloc[:, np.arange(1, 4)]
+        )
```
leewujung, line 284: why do you need np.arange(1,4)? it seems a weird hard-coded thing. It seems that `age1_proportions` only has 3 columns from `age1_metric_proportions`?

echopop/biology.py
```diff
@@ -0,0 +1,1212 @@
...
+        "aged_weight_proportions_df": aged_overall_df,
+        "unaged_weight_proportions_df": unaged_weight_proportions_df.reset_index(),
+        "aged_unaged_sex_weight_proportions_df": (
+            aged_unaged_sex_proportions.astype(float).reset_index().fillna(0.0)
+        ),
+        "aged_unaged_weight_proportions_df": aged_unaged_proportions,
+    }
+
+
+def age1_metric_proportions(
```
leewujung, line 791: Would be good to add docstrings for this and the below functions.

echopop/analysis.py
```diff
@@ -0,0 +1,600 @@
...
+    transect_distance,
+)
+from .statistics import stratified_transect_statistic
+
+
+def process_transect_data(
+    input_dict: dict, analysis_dict: dict, settings_dict: dict, configuration_dict: dict
+) -> dict:
+    """
+    Calculates stratified mean statistics for a set of transects
```
leewujung, line 38: This seems to a result of copy-pasting? Would be good to write out what the outputs are since "process" is a pretty general term.

echopop/analysis.py
```diff
@@ -0,0 +1,600 @@
...
+        A dictionary containing dataframes that define the KS and INPFC stratum limits.
+    settings_dict: dict
+        Dictionary that contains all of the analysis settings that detail specific algorithm
+        arguments and user-defined inputs.
+    """
+
+    # Toggle the correct transect data
+    # ---- The 'original' acoustic transect data
+    if settings_dict["dataset"] == "transect":
+        # ---- Define the and prepare the processed and georeferenced transect data
```
leewujung, line 215:
```suggestion
        # ---- Define and prepare the processed and georeferenced transect data
```

echopop/spatial/transect.py
```diff
@@ -0,0 +1,332 @@
...
+    # ---- Change strata values
+    transect_info = transect_info[[new_stratum, "transect_spacing"]]
+    # ---- Select analysis variable
+    transect_info[f"{settings_dict['variable']}"] = transect_data[f"{settings_dict['variable']}"]
+
+    # Return the output
+    return transect_info.reset_index()
+
+
+def transect_distance(transect_data: pd.DataFrame):
```
leewujung, line 103: I kinda prefer `transect_spatial` or `transect_spatial_features` over `transect_distance` for the function name

echopop/analysis.py
```diff
@@ -0,0 +1,600 @@
...
+    # ---- Age-1 abundances (unaged fish)
+    analysis_dict["biology"]["population"]["tables"]["abundance"].update(
+        {"unaged_age1_abundance_df": unaged_age1_abundance}
+    )
+
+    # Return the updated analysis attribute and `biomass_summary`
+    return biomass_summary, analysis_dict
+
+
+def stratified_summary(
```
leewujung, line 193: It's excellent that you made it work for both the transect and kriging results!

echopop/analysis.py
```diff
@@ -0,0 +1,600 @@
...
+    # Stratified the kriging mesh
+    kriged_results["mesh_results_df"] = stratify_mesh(
+        input_dict, kriged_results["mesh_results_df"], settings_dict
+    )
+
+    # Return kriged (interpolated) results
+    return kriged_results, analysis_dict
+
+
+def apportion_kriged_values(
```
leewujung, line 332: This is a super long function -- I read through it at high level and think it is well done!
One suggestion: How about factoring out 1) the part related to imputation, and 2) the part about Additional reapportionment if age-1 fish are excluded? These seem to be fairly standalone, and if you factor them out, you can forget about the details when reading this function.

echopop/biology.py
```diff
@@ -0,0 +1,1212 @@
...
+        "age_proportions_df": age_number_proportions,
+        "sex_proportions_df": sex_number_proportions,
+        "aged_length_proportions_df": aged_number_proportion,
+        "unaged_length_proportions_df": unaged_number_proportion,
+    }
+
+
+def fit_length_weights(proportions_dict: dict, length_weight_dict: dict) -> pd.DataFrame:
+    """
+    Calculate the average weight per length-bin across all and sexed fish
```
leewujung, line 356:
```suggestion
def fit_length_weights(proportions_dict: dict, length_weight_dict: dict) -> pd.DataFrame:
    """
    Calculate the weight proportion for each length-bin across all and sexed fish
```

echopop/biology.py
```diff
@@ -0,0 +1,1212 @@
...
+        * weight_fitted_df["length_bins"] ** weight_fitted_df["rate"]
+    )
+    # ---- Drop unused columns
+    weight_fitted_df = weight_fitted_df.filter(
+        ["length_intervals", "species_id", "sex", "weight_modeled"]
+    )
+
+    # Adjust for cases where there are too few (< 5) specimens within a given length bin
+    # ---- Count number of specimens across length bins
+    weight_fitted_distribution_df = specimen_data_all.count_variable(
```
leewujung, line 97: This is not really the right place to mark this, but since `count_variable` is not part of the changed code so I can't comment on it easily.
I wonder if what you are trying to do in `count_variable` is what [pandas groupby value_counts](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.value_counts.html) does

echopop/statistics.py
```diff
@@ -0,0 +1,268 @@
...
+    Parameters
+    ----------
+    values: np.ndarray
+        An array of values
+
+    Notes
+    -----
+    This function calculates the 95% confidence interval (assuming a Normal) distribution
+    for the bootstrapped stratified samples. This is done as opposed to using the percentile
+    method for estimate the intervals.
```
leewujung, line 266: I wonder why we can assume normal here? Is that the underlying assumption for what the stratified distributions would follow?

leewujung (approved):
Hey @brandynlucca : Thanks for the PR!

This contains a lot of changes and I am running out of time, so I focused on high-level review for the main modules directly under `echopop/echopop`, and `utils/operations.py` -- the latter is because some functions from that were called and I felt weird. I was not able to review what's under `spatial`.

I think the PR is in good shape to be merged. Most of the things I checked look great.
I like the grouping you have in `analysis.py` -- good job!
I may have small suggestions for the method names in `Survey`, but they do make sense now so it's fine just keep them as is.

For some utility functions that you have in `operations.py` and `statistics.py`: I usually would air on the side of avoiding over-refactor. Things like `operations.bin_value` and `statistics.array_math` are only used twice... is that necessary? 

I think what's going to be important in the next release is to make sure the implementations here corresponds with what we have in the math expressions in the docs. We'll need to figure out a good structure to do it.

Also I am not sure why some tests are commented out. It will be good to add the major ones back before release, so that you know things are solid and ready to go. I'll leave it to you to which ones you want to revive before release. We can pick up the rest in the next release.

docs/example_notebooks/example_echopop_workflow.ipynb
```diff
@@ -0,0 +1,2325 @@
+{
```
leewujung, line 1: <p>This is an interesting solution that you have!</p><p> </p><p>You can achieve the same effect via admonitions supported by MyST.</p><p><a href="https://jupyterbook.org/en/stable/content/content-blocks.html#notes-warnings-and-other-admonitions" target="_blank">https://jupyterbook.org/en/stable/content/content-blocks.html#notes-warnings-and-other-admonitions</a></p><p> </p><p>Same with the TIP block.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/OSOceanAcoustics/echopop/pull/241/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

brandynlucca: <p>Oh, that is much easier than what I was implementing! </p>

echopop/analysis.py
brandynlucca: I'm somewhat unclear on what text in this line seems to be the result of copy-pasting? I agree on the latter point though!

echopop/acoustics.py
brandynlucca: Here I was just trying to replicate the `EchoPro` code that explicitly rounds the acoustically estimated $\rho_{N}$ values. Otherwise, I'd agree with your statement. In general, I think that the differences produced by the choice to round to the closest whole individual (per unit area) versus not are extremely small. Moreover, whatever amount of uncertainty is introduced with either treatment is going to be massively outsized by the uncertainty introduced from the calculation of mean $\sigma_{bs}$ for each stratum.

echopop/analysis.py
leewujung: "Calculates stratified mean statistics for a set of transects" ? clearly not the intent of this function?

echopop/acoustics.py
leewujung: yeah, I agree the difference would be minuscule. But since this is the case, how about just removing that `np.round`? It is more straightforward (and correct IMHO) without it.

echopop/acoustics.py
brandynlucca: I see your point since both this function the (previously known called) `summarize_sigma_bs(...)` can just as easily applied via `.groupby[["other contrasts", "species_id"]]` rather than have those ingrained within the individual functions. I suppose that results in one less thing to check for in terms of bugs, errors, etc?

echopop/acoustics.py
brandynlucca: There are four columns: 1) stratum, 2) number proportion, 3) weight proportion, and 4) NASC proportion. Those are computed for all columns. Perhaps there is a more sophisticated way to map out applying the same simple subtraction to all three columns since `np.arange(1, 4)` at least yields the correct column numbers (`array([1, 2, 3]`) that should always remain constant regardless of the number of strata.

echopop/statistics.py
brandynlucca: Mostly for the sake of simplicity since the default number of replicates is 10,000, and typically it's assumed that (boostrapped) stratified samples will be asymptotic Gaussian although that is not always the case (nor a required assumption). That being said, I have implemented a few additional parameteric and non-parametric methods for computing the CI's (which will be useful way down the road when we formally incorporate bootstrapping at a later release). The augmented code now  defaults to the Bias-Corrected and Accelerated (BCa) non-parameteric method to compute the bootstrap intervals (distribution skewness is estimated via jackknife to compute the acceleration parameter). In cases where the distributions are too skewed and therefore outside the scope of the BCa, the implementation reverts to using the studentized intervals with jackknife bias-correction.

echopop/acoustics.py
leewujung: I suggest that you use column names instead of indices, to make this more robust to errors.

docs/core_data_structure.md
```diff
@@ -1,6 +1,186 @@
...
+
+There are currently **five** unique class attributes that store different types of information and data. 
+
+* `Survey.meta`: Necessary information such as the date the object was created and general data workflow/provenance would be collected.
+* `Survey.config`: Background configuration settings. 
+* `Survey.input`: Imported acoustic, biological, kriging, and stratification [data detailed elsewhere](input-files).
+* `Survey.analysis`: Intermediate data products, variables, and other calculations that may be both relevant to users (e.g. number proportions) or are required for computing the final results. 
+* `Survey.results`: Overall results for each analysis.
+
+Each of these attributes are organized using nested dictionaries that may contain an uneven number of levels. For instance, calculated number and weight proportions computed via the transect analysis are nested via the following structure:
```
leewujung, line 16:
```suggestion
Each of these attributes are organized using nested dictionaries that may contain different number of levels. For instance, number and weight proportions computed via the `transect_analysis` method are nested in the following structure:
```

docs/core_data_structure.md
```diff
@@ -1,6 +1,186 @@
...
+* Survey object
+  * `analysis` attribute
+    * `transect` key
+      * `biology` sub-key
+        * `proportions` sub-key
+          * `number` sub-key
+            * Various number proportion `DataFrame`'s
+          * `weight` sub-key
+            * Various weight proportion `DataFrame`'s
```
leewujung, line 25: I would suggest removing the attribute and sub-key specifiers to avoid confusion. These are all dictionary keys all the down to each of the "leaves", and as you know, key-value has very specific meaning in python.

docs/core_data_structure.md
```diff
@@ -1,6 +1,186 @@
...
+which is equivalent to the following Python code:
```
leewujung, line 26:
```suggestion
which can be accessed using the following Python code:
```

docs/core_data_structure.md
```diff
@@ -1,6 +1,186 @@
 (data-structure)=
 # Core data structures
 
-The whiteboard image below captures the planned core data structure of the `Survey` object that is central to Echopop. Most of the structure is in place but we will fully implement them very soon (tracked in issue [#169](https://github.com/OSOceanAcoustics/echopop/issues/169)).
+Input data, intermediate variables and calculations, and results are stored within several attributes associated with the `Survey`-class object. Information on how these attributes are used in the general `echopop` workflow can be found in greater detail [within the example notebook](echopop-example-workflow).
```
leewujung, line 4: This is ok for now, but I think we should have a diagram or table on the top of this page as a quick reference. I'll add an issue to track this.

docs/core_data_structure.md
```diff
@@ -1,6 +1,186 @@
...
+- `haul_to_transect_df`: Haul-to-transect key that links haul numbers to their respective transects.
+- `length_df`: Unaged (Station 1) length measurements.
+- `specimen_df`: Aged (Station 2) length and weight measurements.
+
+#### Spatial data (`'[spatial]'`)
+- `strata_df`: The length-based 'KS' stratum definitions and fraction of hake for each haul.
+- `geo_strata_df`: The latitudinal extents/ranges of each KS stratum throughout the survey region.
+- `inpfc_strata_df`: The INPFC stratum definitions and their respective latitudinal limits
+
+#### Statistical data (`'[statistics]'`)
```
leewujung, line 98:
```suggestion
#### Statistical data (`['statistics']`)
```

docs/core_data_structure.md
leewujung: Oh I had missed what you have as a TOC since it didn't get rendered in GH, but it got rendered in RTD -- awesome! I would just suggest some edits here then!
```suggestion
Input data, intermediate variables and calculations, and results are stored within several attributes associated with the `Survey`-class object. The attributes and stored data are detailed in the sections below. See [the example notebook](echopop-example-workflow) for information on how these attributes are used in the general `echopop` workflow. 
```

leewujung (approved):
Thanks for the changes. It's cool to see this all coming together! I think this is ready to be merged now. 🎉

Issue #125, opened 2023-12-02
Create a config file validator


leewujung:
Right now the config files are just read as is and the only check is checking the file existence. It will be good to have a validator to check the content of the file.

This came from a TODO in `_check_init_file`:
https://github.com/uw-echospace/EchoPro/blob/b8314f13b812272738a10344034f3c19977baf90/EchoPro/survey.py#L114-L116

and a similar TODO in `_check_survey_year_file`:
https://github.com/uw-echospace/EchoPro/blob/b8314f13b812272738a10344034f3c19977baf90/EchoPro/survey.py#L138-L140

Below from #13:
> - The configuration files have all necessary variables defined and these values are of the correct type. Here we could also check that the paths provided exist.

brandynlucca:
The second `TODO` item has been addressed:

https://github.com/OSOceanAcoustics/echopop/blob/f244105852a86a7ea0fa887401321b39106e5c90/echopop/utils/data_file_validation.py#L5-L66

The first `TODO` item remains at-large.

PR #248, opened 2024-06-03
Add configuration YAML file validator

brandynlucca:
This relates to Issue #125 and implements a new function (`validate_config_structure`) that iteratively parses through the imported entries from the configuration YAML files to assess 1) the correct key names and 2) that the values/inputs are of the expected datatypes.

### DIFF
--- a/echopop/core.py
+++ b/echopop/core.py
@@ -12,7 +12,7 @@ CONFIG_INIT_API = {
     "stratified_survey_mean_parameters": {
         "strata_transect_proportion": float,
         "num_replicates": int,
-        "mesh_transects_per_latitude": int
+        "mesh_transects_per_latitude": int,
     },
     "bio_hake_len_bin": [float],
     "bio_hake_age_bin": [float],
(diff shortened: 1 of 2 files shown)


echopop/utils/load.py
```diff
@@ -575,3 +578,58 @@ def prepare_input_data(input_dict: dict, configuration_dict: dict):
...
+                )
+            for index, item in enumerate(data):
+                validate_value(item, spec[0], f"{path}[{index}]")
+        else:
+            raise ValueError(f"Unknown spec type at {path}: {spec}")
+
+    # Helper function for assessing the data type of each nested value
+    def is_valid_type(data, spec_type):
+        if spec_type is float:
+            return isinstance(data, (int, float))
```
leewujung, line 629: I think a better handling here is probably to coerce the conversion from `int` to `float` and then spit out a warning to let user know about it. This check here can lead to bugs that are hard to check, because you are not checking the type as it is written.

echopop/utils/load.py
```diff
@@ -575,3 +578,58 @@ def prepare_input_data(input_dict: dict, configuration_dict: dict):
...
+
+def validate_config_structure(yaml_data, config_spec):
+    """
+    Validate configuration YAML dictionary structure and entry types
+    """
+
+    # Helper function for validating dictionary key names/structure
+    def validate_dict(data, spec, path=""):
+        for key, value in spec.items():
+            current_path = f"{path}.{key}" if path else key
```
leewujung, line 591: I was confused by the name "path" here. How about `current_branch`?

leewujung (approved):
Hey @brandynlucca : Thanks for the PR! I think it is ready to go. I had 2 small comments that are hopefully straightforward.

I think that the validation can leverage [pydantic](https://docs.pydantic.dev/latest/), but since you already have this working, let's not worry about doing it. I added an issue #252 for using pydantic for validation at some point.

echopop/core.py
```diff
@@ -7,6 +7,71 @@
 import numpy as np
 import pandas as pd
 
+# Required configuration initialization YAML structure
+CONFIG_INIT_API = {
```
leewujung, line 11: I feel using API here is a bit strange. How about we borrow the pydantic lingo to use "model"?
```suggestion
CONFIG_INIT_MODEL = {
```

echopop/core.py
```diff
@@ -7,6 +7,71 @@
...
+    "kriging_parameters": {
+        "A0": float,
+        "longitude_reference": float,
+        "longitude_offset": float,
+        "latitude_offset": float,
+    },
+}
+
+# Required data configuration YAML structure
+CONFIG_DATA_API = {
```
leewujung, line 37:
```suggestion
CONFIG_DATA_MODEL = {
```

PR #251, opened 2024-06-18
Semivariogram parameter fitting/optimization

brandynlucca:
A new `Survey`-class method (`fit_variogram(...)`) has been added to allow for the non-linear least squares fitting of variogram parameters. This includes options for only fitting select parameters (instead of all), as well as modifying different parameter values (i.e. `initial_values`), bounds (i.e. `lower_bounds`, `upper_bounds`), and various arguments specific to the optimization methods/algorithm. These results can then be propagated over to the `Survey.kriging_analysis(..., best_fit_variogram = True)` method.


echopop/survey.py
```diff
@@ -213,20 +216,299 @@ def stratified_analysis(
...
+            optimization_parameters = {
+                "max_fun_evaluations": max_fun_evaluations,
+                "cost_fun_tolerance": cost_fun_tolerance,
+                "solution_tolerance": solution_tolerance,
+                "gradient_tolerance": gradient_tolerance,
+                "finite_step_size": finite_step_size,
+                "trust_region_solver": trust_region_solver,
+                "x_scale": x_scale,
+                "jacobian_approx": jacobian_approx,
+            }
```
leewujung, line 461: I think we can make the input argument list less daunting by requiring users to put in a dict for updating any of the default params. If I remember correctly it is not recommended to default an input argument to a dict directly (because it is mutable?), but we can use a preset dict for that. Something like:
```python
DEFAULT_OPT_DICT = {
    "param1": value1, ...
}

... # code blocks in the middle

# assumes opt_params below is the user input argument
opt_params_out = DEFAULT_OPT_DICT
if not isinstance(opt_params, dict):  # make sure input is a dict
    raise ValueError("this should be a dict")
if opt_params:  # update using all input params at once
    opt_params_out.update(opt_params)
```

echopop/survey.py
```diff
@@ -213,20 +216,299 @@ def stratified_analysis(
...
+            variogram_parameters["lag_resolution"] = lag_resolution
+        # ---- Max range
+        if max_range is not None:
+            variogram_parameters["range"] = max_range
+        # ---- Number of lags
+        variogram_parameters["n_lags"] = n_lags
+        # ---- Azimuth range
+        variogram_parameters["azimuth_range"] = azimuth_range
+        # ---- Force lag-0
+        variogram_parameters["force_lag_zero"] = force_lag_zero
```
leewujung, line 447: I think we do something similar to [what i commented on on the optimization params](https://github.com/OSOceanAcoustics/echopop/pull/251/files#r1720595160), to require users to set the parameters in a dict using key-value pairs, have the default values in a dict to initialize `variogram_parameters` first, and then update `variogram_parameters` with user inputs via the `.update` function. Like:
```python
DEFAULT_VPARAM_DICT = {
    "param1": value1, ...
}

... # code blocks in the middle

# assumes variogram_params below is a user input argument
v_params_out = DEFAULT_VPARAM_DICT
v_params_out["model"] = model
if not isinstance(variogram_params, dict):  # make sure input is a dict
    raise ValueError("this should be a dict")
if variogram_params:  # if not None
    # update using all input params at once
    v_params_out.update(variogram_params)
```

echopop/analysis.py
```diff
@@ -243,6 +252,106 @@ def stratified_summary(
...
+        # ---- Skip file dir in message
+        _warn_skips = (os.path.dirname(__file__),)
+        # ---- Generate warning
+        warnings.warn(
+            f"Default `range` parameter ({variogram_parameters['range']}) does not "
+            f"align with the inputs for `lag_resolution` ({lag_resolution}) "
+            f"and `n_lags` ({n_lags}). The range will be changed to {max_range}.",
+            stacklevel=2,
+            skip_file_prefixes=_warn_skips,
+        )
```
leewujung, line 292: I am confused by this. From the if statement, it seems that unless `variogram_parameters["range"]` is set to `max_range`, it will be reset to `max_range` and a warning will be raised. Wouldn't this be the same as not having `range` in the `variogram_parameters` dictionary  and just compute it from `n_lag * lag_resolution`? Did I miss something here?

echopop/spatial/variogram.py
```diff
@@ -384,5 +901,421 @@ def variogram(
...
+    # ---- Parameter values
+    optimization_values = [
+        max_fun_evaluations,
+        cost_fun_tolerance,
+        solution_tolerance,
+        gradient_tolerance,
+        finite_step_size,
+        trust_region_solver,
+        x_scale,
+        jacobian_approx,
```
leewujung, line 1056: I think you can omit this by passing in `optimization_parameters` as a dict instead of using ** to expand it at input of this function (`create_optimization_options`).

echopop/spatial/variogram.py
```diff
@@ -384,5 +901,421 @@ def variogram(
...
+    ]
+    # ---- Drop unused
+    unused_parameters_bitmap = [k is not None for k in optimization_values]
+    used_parameters = [
+        value for bit, value in zip(unused_parameters_bitmap, optimization_parameters) if bit
+    ]
+    used_values = [
+        value for bit, value in zip(unused_parameters_bitmap, optimization_values) if bit
+    ]
```
leewujung, line 1065: following from the above: If you have `optimization_parameters` as a dictionary, you can `pop` the items with values equal to `None`.

echopop/survey.py
```diff
@@ -213,20 +216,299 @@ def stratified_analysis(
...
+            Threshold used for determining convergence via incremental changes of the cost function.
+            Defaults to 1e-6.
+        solution_tolerance: float
+            Threshold used for determining convergence via change of the independent variables.
+            Defaults to 1e-8.
+        gradient_tolerance: float
+            Threshold used for determining convergence via the gradient norma. Defaults to 1e-8.
+        finite_step_size: float
+            The relative step sizes used for approximating the Jacobian via finite differences.
+        trust_region_solver: Literal["exact", "float"]
```
leewujung, line 355: Is the suggested change below right?
```suggestion
        trust_region_solver: Literal["exact", "base"]
```

leewujung (commented):
Hey @brandynlucca : thanks for the PR, good work! exciting to see the fitting functionality! 

I have some inline comments, mostly for bundling things up so that the code would look a bit cleaner and easy to follow.

I read through all the main variogram functions called by `Survey.fit_variogram` but not the various variogram models.

Because I didn't read the variogram models, I also noticed that they do not have accompanying tests. I think we can do without them in this PR, but it would be important to add them to make sure that the calculations are implemented correctly. I'll create an issue for this and let's discuss how to tackle it more efficiently.

PR #255, opened 2024-06-22
Fix/organize f-strings in summary outputs

brandynlucca:
This includes small adjustments to the `message` outputs that generate the console results when running methods like `transect_analysis` when `verbose = True`.


leewujung (approved):
LTGM!

PR #257, opened 2024-06-25
Raw Echoview export ingestion and batch processing

brandynlucca:
This PR includes a variety of new features for ingesting and batch processing Echoview exports to both 1) write a new `xlsx` file with the consolidated vertically integrated (NASC) exports for age-1+ and age-2+ datasets and 2) directly update the `Survey` class object to read the updated data. These changes include:

**New batch loader sub-module**
* The `batch_load` sub-module has been added to the `utils` sub-package. This is called when the `Survey`-class object is initialized, which adds additional arguments to the initialization. Alternatively, the primary batch processing function, `bach_read_echoview_exports(...)`, can be invoked independent of the `Survey` object.

**Data parameterization changes**
* A new section in `initialization_config.yml` (`nasc_exports`) introducing various keys for parameterizing the outputs.
* A new section in `survey_year_****_config.yml` enabling filepaths to echogram regions (i.e. `export_regions`).
* Two new mapping keys have been added to `core.py` required for data validation (`ECHOVIEW_EXPORT_MAP` and `REGION_EXPORT_MAP`).
* Mismatches in age-1+ and age-2+ region IDs, stratum numbers, and haul numbers necessitated a separator similar to `NASC_all_ages` vs `NASC_no_age1`. The `*_all_ages` and `*_no_age1` have now been generalized as appended 'tags' to age group-specific values that are then converted to shared column names (i.e., `nasc`, `stratum_num`, and `haul_num`) depending on the `exclude_age1` argument value. This necessitated some changes to the `load` sub-module in the `utils` sub-package.

**Housekeeping changes**
* Small changes to `.pre-commit-config.yaml` and the creation of an associated `.codespell-ignore-words.txt` that is compatible with misspelled column names in data sources.

**Note**
* There is a small outstanding issue concerning mismatches in the output of the batch loader (which matches the previous `NASC construction` operation in EchoPro) and the pre-generated consolidated export files we have received previously. This is being diagnosed, but appears to be due to differences in the source export files rather than anything within the code.


echopop/spatial/transect.py
```diff
@@ -346,3 +363,130 @@ def transect_bearing(transect_data: pd.DataFrame):
...
+        transect_summary = (
+            transect_data.groupby(index_variable)["max_depth"].max().to_frame("bottom_depth")
+        )
+
+    # Calculate the mean layer depth and height
+    if "layer_depth_min" not in transect_data.columns:
+        raise ValueError("The column 'layer_depth_min' is missing from `transect_data`.")
+    elif "layer_depth_max" not in transect_data.columns:
+        raise ValueError("The column 'layer_depth_max' is missing from `transect_data`.")
+    else:
```
leewujung, line 408: This is fine. One way to do this is since the checks are all for columns, you could check the three first , and then if nothing fails all the computation follows. Something like:
```python
for col in ["max_depth", "layer_depth_min", "layer_depth_max"]:
   if col not in transect_data.columns:
      raise ValueError(f"...")

# If no errors are raised, proceed with computation
# ---- Compute the bottom depth
transect_summary = ...
# ---- Mean layer depth
...
```

This way the code is pretty clean without the many if-else.

echopop/spatial/transect.py
```diff
@@ -346,3 +363,130 @@ def transect_bearing(transect_data: pd.DataFrame):
...
+                mean_depth_min=("layer_depth_min", "min"), mean_depth_max=("layer_depth_max", "max")
+            )
+            .assign(height=lambda x: x.mean_depth_max - x.mean_depth_min)["height"]
+        )
+
+    # Return the output
+    return transect_summary.reset_index()
+
+
+def export_transect_spacing(
```
leewujung, line 430: This loop below obviously gets the job done:
```
# Iterate through the transects to determine the maximum spacing
for i in range(len(transect_number)):
```
I am thinking that this could vectorized by groupby ops.
Since `latitude_threshold` is fixed, you can subset the dataframe first and then groupby based on `transect_num`, and compute mean, max, min from the groups. Then you could get the spacing with the right lag pretty easily (diff offset by 2), and then do the filtering (the if condition at the end) and assign them all.

Not sure how fast/slow the current ops are. If it is slow, this is worth considering.

echopop/survey.py
```diff
@@ -10,7 +10,7 @@
     stratified_summary,
 )
 from .core import DATA_STRUCTURE
-from .utils import load as el, message as em
+from .utils import batch_load as ebl, load as el, message as em
```
leewujung, line 13: I am surprised that black/isort allows this! 🤪

echopop/survey.py
```diff
@@ -44,15 +44,38 @@ class Survey:
...
+        self,
+        init_config_path: Union[str, Path],
+        survey_year_config_path: Union[str, Path],
+        construct_nasc: bool = False,
+        transect_pattern: Optional[str] = r"T(\d+)",
+        index_variable: Union[str, List[str]] = ["transect_num", "interval"],
+        unique_region_id: str = "region_id",
+        region_class_column: str = "region_class",
+        export_file_directory: Optional[Union[str, Path]] = None,
+        export_save_directory: Optional[Union[str, Path]] = None,
```
leewujung, line 56: This would require refactoring (hopefully not too much!?), but seeing the large number of input arguments you added here, I think it would be better/cleaner if we keep the initialization of a Survey object just as initialization, and have the NASC ingestion part in a method. These arguments are really only used for the ingestion and not anywhere else. Separating it out as a method makes it explicit to the users what they are expected to do - to ingest NASC data (ie. not hiding everything under the hood of initialization, especially because it would take some time to compute), and also provide flexibility to skip the ingestion/load previously organized data.

echopop/utils/operations.py
```diff
@@ -239,69 +239,100 @@ def stretch(
...
+
+    # Restore original dtypes of columns in merged_frame
+    for col in merged_frame.columns:
+        dtype_to_restore = unique_dtypes.get(col)
+        if dtype_to_restore is not None:
+            if pd.api.types.is_integer_dtype(dtype_to_restore):
+                # Check if column contains NaN or inf values
+                if merged_frame[col].isnull().any() or not merged_frame[col].notna().all():
+                    continue  # Skip conversion if NaN or inf present
+                merged_frame[col] = merged_frame[col].astype(dtype_to_restore)
```
leewujung, line 334: This seems super specific and would probably cause some issues depending on the data it encounters. I know that NaN cannot exist as an integer, but I guess the question(s) may be:
- if the values need to be integer, why is there NaN in the first place, and can those be 0?
- does it need to be integer, can we use float (1.0, 2.0, etc) so that NaN is retained?

echopop/survey.py
leewujung: Ok, after looking into what each py file have, I would suggest the following:
- rename `batch_load` to `load_nasc` and `load` to `load_survey`
- use the full module name, so `from .utils import load_nasc, load_survey, message` and just have a few more characters in the code when importing functions from these modules.

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+
+    # Get coordinate column names, if present
+    # ---- Extract longitude column name
+    lon_col = next((col for col in export_file.columns if "lon" in col.lower()), None)
+    # ---- Extract latitude column name
+    lat_col = next((col for col in export_file.columns if "lat" in col.lower()), None)
+
+    # Automatic "bad data" removal/imputation for georeferenced (Latitude/Longitude) data
+    # ---- Helper imputation function
+    def impute_bad_data(data, column):
```
leewujung, line 46: Is there a reason why you don't want to use [pd.dataframe.fillna](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html) for these?

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+        # ---- Join files on multiple lines
+        unparsed_transects_formatted = "\n".join(unparsed_transects)
+        if len(unparsed_transects) > 10:
+            # ---- Get number of unparsed strings
+            unparsed_n = len(unparsed_transects)
+            # ---- Raise error
+            raise ValueError(
+                f"Transect numbers could not be parsed from {unparsed_n} files in {file_directory}."
+                f" The first 10 files include:\n {unparsed_transects_formatted}"
+            )
```
leewujung, line 175: This will still print out all of the files, right? since you did not do any selection in `unparsed_transects_formatted = "\n".join(unparsed_transects)`.

Why not just get those into a list and print the list, and let the IDE do the job of cutting off/hiding the outputs by their rules?

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+        # ---- Create key-specific filter pattern
+        pattern = "|".join(
+            [r"(?<!-)\b{}\b".format(re.escape(name.lower())) for name in region_names[key]]
+        )
+        # ---- Apply filter to regions dataframe
+        new_region_filtered = new_region_df[
+            new_region_df["region_class"].str.contains(pattern, case=False, regex=True)
+        ]
+        # ---- Concatenate to the initialized/pre-existing DataFrame
+        region_df = pd.concat([region_df, new_region_filtered], ignore_index=True)
```
leewujung, line 294: I think it is more memory-efficient/faster if you accumulate a list and then use `pd.concat(that_list)` to turn that into a dataframe at once.

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+            f"The defined `region_class_column` ({region_class_column}) used for applying "
+            f"`region_filter`does not exist!"
+        )
+    # ---- Convert to list, if needed
+    if isinstance(region_filter, str):
+        region_filter = list(region_filter)
+    elif not isinstance(region_filter, list):
+        raise TypeError(
+            f"The defined `region_filter` ({region_filter}) must be either a `str` or `list`."
+        )
```
leewujung, line 364: Seems that the isinstance check for both list and str should happen first, and then the conversion from str to list happens if necessary.

leewujung: Same with the `index_variable` check below.

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+    # ---- Validate existence
+    if not Path(save_folder).exists():
+        raise FileNotFoundError(f"Save directory for NASC file ({save_folder}) does not exist.")
+
+    # Check that directory exists
+    # ---- Get export file directory
+    export_file_directory = export_settings["export_file_directory"]
+    # ---- Drop prepended "/" or "\\" to avoid using an absolute path
+    if export_file_directory.startswith("/") or export_file_directory.startswith("\\"):
+        export_file_directory = export_file_directory[1:]
```
leewujung, line 483: I think it may be a good idea to not leave this "back door" open for users to be able to separately put in `file_directory` and `save_directory`, and force all operations to be via the config files. This way we can always (at least within reasonable ranges) trace back on which config file ingest what data and export what files. So essentially I think **less** flexibility here is better, so that we prevent accidental shorthands or things people forget to change.

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+            raise TypeError(f"Input directory {file_directory} must either be a `Path` or a `str`.")
+    else:
+        file_directory = str(Path(configuration_dict["data_root_dir"]) / export_file_directory)
+
+    # Validate
+    # ---- Check if directotry exists
+    if not Path(file_directory).exists():
+        raise FileNotFoundError(f"The export file directory {{{file_directory}}} not found!")
+    # ---- Check if there are any files
+    elif not any(Path(file_directory).iterdir()):
```
leewujung, line 500: this should just be an `if`

echopop/utils/batch_load.py
```diff
@@ -0,0 +1,608 @@
...
+            full_interval_strata_df[unique_region_id].fillna(999).astype(int)
+        )
+        # ---- Fill haul with 0's
+        full_interval_strata_df["haul_num"] = (
+            full_interval_strata_df["haul_num"].fillna(0).astype(int)
+        )
+        # ---- Fill stratum with 1's
+        full_interval_strata_df["stratum_num"] = (
+            full_interval_strata_df["stratum_num"].fillna(1).astype(int)
+        )
```
leewujung, line 575: Just wondering why you fill missing stratum with 1s. Is it because stratum=1 does not get processed?

leewujung (commented):
Hey @brandynlucca : Sorry it took me so long to get to this. Most of comments are small, but there are 2 bigger ones:
- Separating Echoview NASC ingestion out from the Survey class initializer: [this comment](https://github.com/OSOceanAcoustics/echopop/pull/257/files#r1712205250)
- Not allowing a backdoor to set the source and export locations for the batch processing function, and make it strictly/only via the config file: [this comment](https://github.com/OSOceanAcoustics/echopop/pull/257/files#r1712363571)

Feel free to ping me on slack if you want to discuss!

echopop/survey.py
brandynlucca: > I am surprised that black/isort allows this! 🤪
I actually believe that `black` recommended this change (I could be misrembering...).

echopop/utils/batch_load.py
brandynlucca: `stratum=1` gets processed. This just comports with what Chu had written for `EchoPro`, so I wanted to maintain that consistency across files.

echopop/utils/batch_load.py
brandynlucca: I have tried implementing this by created an appropriately second indexed `DataFrame` to fill in the values, but that didn't seem to match what Chu's code produced. The unfortunate thing here is that these are rolling imputations, so they are conditional on the previously imputed values. I guess an alternative would be to try and compute the slopes between coordinates and impute that way; however, I tried doing that previously and it didn't seem to work out. That could have very well been simply due to an indexing issue on my end.

echopop/utils/operations.py
brandynlucca: This is necessary due to how `pandas.merge()` converts `int` to `float` when there are potentially mismatched records (i.e. exist in one but not another DataFrame) anywhere. So this affects the entire DataFrame rather than specific columns where there may be disagreement/conflict/contain `NaN` or `inf`. Where `NaN` may occur are other cases where there is some mismatch in the merge keys (or having a missing value to begin with) across dataframes. But `int` values needs to remain `int` to preserve various indexing functions that occur later on in the package.

echopop/survey.py
brandynlucca: This would require either then moving `load_survey_data` out as well as a completely separate method (i.e. `Self.load_survey_data`) outside of the initialization, or shoehorning it into `Survey.transect_analysis`. This is because components of `load_survey_data` are contingent upon output files (i.e. the haul-transect key file) generated by the `batch_read_echoview_exports` function. 

> self.analysis["settings"].update(...) can be done in the method too.

I am a little unclear on what you are referring to here.

brandynlucca: Per discussion:
* Separate method for `batch_read_echoview_exports`
* Refactor `el.load_survey_data` as a separate method

echopop/survey.py
leewujung: Ok! Good to know!

PR #264, opened 2024-08-28
LiveSurvey Class

brandynlucca:
Draft PR for the `LiveSurvey` class rough draft. 

The workflows that occur: 

```

### CODE REVIEW

File: echopop/live/live_data_loading.py [hunk 1]
```diff
@@ -0,0 +1,928 @@
...
+    # Drop incomplete datasets
+    if dataset == "biology":
+        data_files = validate_complete_biology_dataset(
+            data_files, directory_path, file_configuration
+        )
+
+    # Query the SQL database to process only new files (or create the db file in the first place)
+    valid_files, file_configuration["database"][dataset] = query_processed_files(
+        database_root_directory, file_settings, data_files
+    )
```
leewujung, line 322: This section deviates from the goal to validate paths as suggested by the function name.
Probably better to separate these out.

echopop/live/live_survey.py
```diff
@@ -0,0 +1,343 @@
...
+        return self.__repr__()
+
+    def load_acoustic_data(
+        self, xarray_kwargs: dict = {}, input_filenames: Optional[list] = None, verbose: bool = True
+    ):
+
+        # Validate the data directory and format the filepaths
+        acoustic_files = eldl.validate_data_directory(
+            self.config, dataset="acoustics", input_filenames=input_filenames
+        )
```
leewujung, line 127: Note to self: only new files are in the returned `acoustic_files`, because `validate_data_directory` right now contains `query_processed_files` that would exclude files that have already been processed.

echopop/live/live_survey.py
```diff
@@ -0,0 +1,343 @@
...
+            prc_nasc_df, acoustic_data_units = eldl.read_acoustic_files(
+                acoustic_files, xarray_kwargs=xarray_kwargs
+            )
+            # ---- Add the `acoustic_data_units` to the dictionary
+            self.config["acoustics"]["dataset_units"] = acoustic_data_units
+            # ---- Preprocess the acoustic dataset
+            # TODO: SettingWithCopyWarning:
+            self.input["acoustics"]["prc_nasc_df"] = preprocess_acoustic_data(
+                prc_nasc_df.copy(), self.input["spatial"], self.config
+            )
```
leewujung, line 142: These additions are ad-hoc and deviates from the validated config model.

echopop/live/live_survey.py
```diff
@@ -0,0 +1,343 @@
...
+            # ---- Add meta key
+            self.meta["provenance"].update(
+                {
+                    "acoustic_files_read": acoustic_files,
+                }
+            )
```
leewujung, line 148: These additions are ad-hoc and deviates from the idea behind a validated config model.

echopop/live/live_data_loading.py
```diff
@@ -0,0 +1,928 @@
...
+            if "datetime" in dataframe_combined.columns:
+                dataframe_combined["datetime"] = convert_datetime(dataframe_combined["datetime"])
+            # ---- Add to the data dictionary
+            biology_output[f"{dataset}_df"] = dataframe_combined
+
+    # Return the output
+    return biology_output
+
+
+def read_acoustic_zarr(file: Path, config_map: dict, xarray_kwargs: dict = {}) -> tuple:
```
leewujung, line 201: Suggest changing this to `nasc_zarr_to_df` so that it is clear what it does. Also to avoid the similarity with `read_acoustics_files` that has acousticS and here the function name does not have an S...

echopop/live/live_core.py
```diff
@@ -0,0 +1,387 @@
...
+            "required_keys": ["transmit", "TS_length_regression_parameters"],
+            "optional_keys": [],
+            "keys": {
+                "transmit": {
+                    "required_keys": ["frequency", "units"],
+                    "optional_keys": [],
+                    "keys": {
+                        "frequency": float,
+                        "units": ["Hz", "kHz"],
+                    },
```
leewujung, line 48: I feel this is making the options too liberal...

echopop/live/live_acoustics.py
```diff
@@ -0,0 +1,337 @@
...
+        survey_data.loc[:, "frequency_nominal"],
+        transmit_settings,
+        acoustic_analysis_settings["dataset_units"]["frequency"],
+    )
+    # ---- Filter out any unused frequency coordinates
+    prc_nasc_df_filtered = (
+        survey_data[survey_data["frequency_nominal"] == transmit_settings["frequency"]]
+        # ---- Drop NaN/NaT values from longitude/latitude/ping_time
+        .dropna(subset=["longitude", "latitude", "ping_time"])
+    )
```
leewujung, line 53: I think it will be more efficient to do this selection in `read_acoustic_files` so that what you're doing in this function will be strictly related to spatial stuff (and then the function name can be made more specific also).

echopop/live/live_spatial_methods.py
```diff
@@ -0,0 +1,665 @@
...
+
+    # Convert to a GeoDataFrame and return the GeoDataFrame
+    return gpd.GeoDataFrame(
+        data=bound_df,
+        geometry=gpd.points_from_xy(bound_df["x"], bound_df["y"]),
+        crs=projection,
+    )
+
+
+def apply_griddify_definitions(dataset: pd.DataFrame, spatial_config: dict):
```
leewujung, line 164: I think the first half of `apply_griddify_definitions` (up to # Convert to GeoDataFrame where you actually handles the acoustic data) could be separate out as a function that you run once in the init and store the output grids needed to "assign" the NASC data into.

echopop/live/live_data_loading.py
```diff
@@ -0,0 +1,928 @@
...
+    # ---- Extract all of the file ids
+    biology_config_ids = list(biology_file_ids.keys())
+    # ---- Initialize the dictionary that will define this key in the `input` attribute
+    biology_output = {f"{key}_df": pd.DataFrame() for key in biology_config_ids}
+    # # ---- Create filepath object
+    if "data_root_dir" in file_configuration:
+        # directory_path = Path(file_configuration["data_root_dir"]) / file_settings["directory"]
+        directory_path = "/".join([file_configuration["data_root_dir"], file_settings["directory"]])
+    else:
+        directory_path = file_settings["directory"]
```
leewujung, line 152: I am curious what the problem with using Path object is. Thinking that `Path("") / file_settings["directory"]` would still give Path(file_settings["directory"]) so seems can eliminate the if-else

echopop/live/live_biology.py
```diff
@@ -0,0 +1,1377 @@
...
+    filter_dict = dict(species_id=species_filter, trawl_partition=trawl_filter)
+
+    # Apply the filter
+    filtered_biology_output = {
+        key: biology_data_filter(df, filter_dict)
+        for key, df in biology_output.items()
+        if isinstance(df, pd.DataFrame) and not df.empty
+    }
+    # ---- Create new data flag
+    file_configuration["length_distribution"] = prepare_length_distribution(file_configuration)
```
leewujung, line 100: This addition of "length_distribution" in `self.config` is probably not a good idea because you already have `self.config["biology"]["length_distribution"]`. Since the config dict is defined upfront, I would suggest against adding undefined keys into it in other parts of the code. I think I made another comment also on this.

echopop/live/live_biology.py
```diff
@@ -0,0 +1,1377 @@
...
+    # ---- Incorporate additional data, if new data are present
+    if filtered_biology_output:
+        # ---- Merge the trawl information and app
+        merge_trawl_info(filtered_biology_output)
```
leewujung, line 104: Why do you delete `biology_dict["trawl_info_df"]` at the end of this function?

echopop/live/live_biology.py
```diff
@@ -0,0 +1,1377 @@
...
+    # ---- Add numeric lower boundary
+    length_bins_df["lower"] = length_bins_df["interval"].apply(lambda x: x.left).astype(float)
+    # ---- Add numeric upper boundary
+    length_bins_df["upper"] = length_bins_df["interval"].apply(lambda x: x.right).astype(float)
+
+    # Return the dataframe that will be incorporated into the biological data attribute
+    return length_bins_df
+
+
+def preprocess_biology_data(biology_output: dict, spatial_dict: dict, file_configuration: dict):
```
leewujung, line 80: I need your help to go through the intention of this function, because it is quite hidden behind the many layers of functions that are very specific. I think there's a balance between reusability and readability that we should discuss. I also have an additional question on why we need a database if for new trawls we are basically just appending (accumulating) new data.

echopop/live/live_data_loading.py
```diff
@@ -0,0 +1,928 @@
...
+    # Iterate through the filename-derived tags and add them to the DataFrame
+    for i in valid_tags:
+        matched_key = LIVE_FILE_FORMAT_MAP[i]
+        df_validated[matched_key["name"]] = matched_key["dtype"](match_obj.group(i))
+
+    # Return the resulting DataFrame
+    return df_validated
+
+
+def infer_datetime_format(timestamp_str: Union[int, str]):
```
leewujung, line 526: I wonder if the patterns you included here are already covered in pandas [guess_datetime_format](https://pandas.pydata.org/docs/dev/reference/api/pandas.tseries.api.guess_datetime_format.html)? Since these seem to be pretty standard ones. 

Also, do you need these because the files from FEAT have different datetime formats? For now this obviously works, but I feel we can communicate with Alicia to see if she could make those uniform.

echopop/live/live_data_loading.py
```diff
@@ -0,0 +1,928 @@
...
+
+    # Concatenate the dataframe component
+    acoustic_data_df = pd.concat([df for df, _ in results_list], ignore_index=True)
+    # ---- Add the `acoustic_data_units` to the dictionary and output the resulting tuple
+    return acoustic_data_df, results_list[0][1] if results_list else None
+
+
+def filter_filenames(
+    directory_path: Path, filename_id: str, files: List[Path], file_extension: str
+):
```
leewujung, line 100: I think we can use fsspec for flexible handling between different file systems. Also, it is a lot of work here to get something that is general across all the files, but the patterns are still pretty specific. For the scenario we are dealing with, I'd say generality is not the first priority, and more hard-coding may actually make the code more readable. Let's discuss this when we meet.

echopop/live/live_acoustics.py
```diff
@@ -0,0 +1,337 @@
...
+    acoustic_db = file_configuration["database"]["acoustics"]
+
+    # Create a copy of the dataframe
+    df = nasc_data_df.copy()
+
+    # Add population-specific columns (specified in the file configuration)
+    # TODO: Add to `yaml` file for configuration; hard-code for now
+    add_columns = ["number_density", "biomass_density"]
+    # ----
+    df[add_columns] = 0.0
```
leewujung, line 306: Just trying to make sure that I understand - these two columns are not supposed to be added here, and they are here as a temporary hardcoded fix?

echopop/live/live_data_processing.py
```diff
@@ -0,0 +1,253 @@
...
+    biology_db = file_configuration["database"]["biology"]
+
+    # Check whether data dictionary is empty
+    if acoustic_dict["nasc_df"] is None or acoustic_dict["nasc_df"].empty:
+        # ---- Print, if verbose
+        if verbose:
+            print("No new processed acoustic data available for processing.")
+    else:
+        # Get related acoustic data
+        acoustic_df = get_nasc_sql_data(acoustic_db, acoustic_dict, unique_columns=unique_columns)
```
leewujung, line 74: This only difference in passing `acoustic_dict` and `biology_dict` here (`acoustic_pipeline`) vs in `biology_pipeline` is goes all the way down to in `query_dataset` the `get_unique_identifiers` uses different input `data_dict`. What does `get_unique_identifiers` do? There's a comment
```python
# Get unique values of each contrast column across the biological datasets
```
in that function but perhaps it is not just biological datasets?

leewujung: Got it now after verbal explanation! :)

echopop/live/live_biology.py
```diff
@@ -0,0 +1,1377 @@
...
+            table_name="strata_summary_df",
+            dataframe=strata_df,
+            id_columns=["stratum"],
+        )
+
+    # Get unique strata values
+    strata_values = np.unique(nasc_biology_data["stratum"]).tolist()
+
+    # Update the table
+    sql_update_strata_summary(
```
leewujung, line 1370: This is called in both `biology_pipeline` and `acoustic_pipeline`. Not sure why this update from `acoustic_db` to `biology_db` needs to happen?

leewujung (commented):
Interim summary from me reading the code - not sure where this should go so maybe just putting it here!

`LiveSurvey.init()`  

`LiveSurvey.load_acoustic_data`
- Turn NASC zarr to dataframe  
- Select the frequency wanted  
- Assign x/y grid and stratum to the NASC dataframe entries 

`LiveSurvey.load_biology_data`
- Combine multiple biological csv files  
- Assign the hauls to specific stratum  
- Insert new data into database and read back the whole dataset as dataframe

`LiveSurvey.process_acoustic_data`
- Operations will be skipped if no new data is present in `input['acoustics']['prc_nasc_df']`  
- `.compute_nasc` calls `.integrate_nasc` to integrate NASC  
- `.integrate_nasc()` does 1) integrate NASC, and 2) calculates echometrics  
- `.format_acoustic_datset()`  
  - Adds `number_density` and `biomass_density` into `nasc_data_df` (but not computing them??)  
  - Adds successfully processed files into the acoustics database  
  - Pulls out the entire combined dataset  
- `.acoustic_pipeline()` and `.biology_pipeline()` basically does the same thing, it’s only the entries updated in the biological estimates database is different:  
  - for `.acoustic_pipeline()`, only the specific grid(s) that is associated with the new NASC entries are updated  
  - for `.biology_pipeline()`, all grids in the stratum for which the haul is located are updated  
  - `.get_average_strata_weights()` computes the average strata weights within each of the acoustic or biology pipeline runs  
  - Flow of calculation:  
    - Compute `number_density` from `nasc` and `sigma_bs_mean`
    - Compute `biomass_density` from `number_density` and `average_weight`  
    - `summarize_strata()` updates the `biology_db` with `number_density` and `biomass_density` mean from the `acoustic_db` -- not sure why this is needed  
    - `update_population_grid`: updates grid.db  
- General flow of ops:  
  - New NASC data: `init()` --> `load_acoustic_data()` --> `process_acoustic_data()` --> `estimate_population()`  
  - New catch haul data: `init()` --> `load_biology_data()` --> `process_biology_data()` --> `estimate_population()`

PR #268, opened 2024-09-06
Translation dictionary for data column names

brandynlucca:
This PR adds a dictionary to `.core` called `NAME_CONFIG` that will translate particular columns in the acoustic, biological, kriging, and stratification datasets. In cases where column names (e.g. `Frequency`) are present, they will be converted into the naming scheme used by `Echopop` (e.g. `Frequency -> length_count`). Otherwise, the names remain the same and continue with being validated as normal.

### DIFF
--- a/echopop/core.py
+++ b/echopop/core.py
@@ -72,6 +72,37 @@ CONFIG_DATA_MODEL = {
     },
 }
 
+# Name configuration dictionary
+NAME_CONFIG = {
+    "Age": "age",
+    "Cell portion": "fraction_cell_in_polygon",
+    "Frequency": "length_count",
+    "haul": "haul_num",
+    "haul end": "haul_end",
+    "haul start": "haul_start",
+    "Haul": "haul_num",
+    "Latitude": "latitude",
+    "Latitude (upper limit)": "northlimit_latitude",
+    "Latitude of centroid": "centroid_latitude",
+    "Length": "length",
+    "Longitude": "longitude",
+    "Longitude of centroid": "centroid_longitude",
+    "strata": "stratum_num",
+    "Sex": "sex",
+    "Ship": "ship_id",
+    "Spacing": "transect_spacing",
+    "Species_Code": "species_id",
+    "Species_Name": "species_name",
+    "Strata Index": "stratum_num",
+    "Stratum": "stratum_num",
+    "Transect": "transect_num",
+    "VL start": "vessel_log_start",
+    "VL end": "vessel_log_end",
+    "wt": "fraction_hake",
+    "Weight": "weight",
+    "Weight_In-Haul": "haul_weight",
+}
+
 # `Survey` object data structure
 CONFIG_MAP = {
     "biological": {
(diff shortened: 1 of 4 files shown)


echopop/core.py
```diff
@@ -72,6 +72,37 @@
...
+    "Length": "length",
+    "Longitude": "longitude",
+    "Longitude of centroid": "centroid_longitude",
+    "strata": "stratum_num",
+    "Sex": "sex",
+    "Ship": "ship_id",
+    "Spacing": "transect_spacing",
+    "Species_Code": "species_id",
+    "Species_Name": "species_name",
+    "Strata Index": "stratum_num",
```
leewujung, line 96: These 2 are indeed different names but represent the same thing, right? Just wanted to make sure that's the case.

brandynlucca: By "these" two do you mean `Strata Index` and `Species Name` as highlighted here? Or do you mean `Strata Index` and `Stratum` (I'm assuming the latter)? If so, then yep, they are the same thing.

echopop/utils/load.py
```diff
@@ -414,17 +427,34 @@ def validate_data_columns(
...
+                if not set(validation_settings.keys()).issubset(set(data_columns)):
+                    missing_columns = set(validation_settings.keys()) - set(data_columns)
+                    raise ValueError(f"Missing columns in the Excel file: {missing_columns}")
+
+                # Get present data column names
+                valid_columns = [
+                    o
+                    for o, t in zip(original_columns, data_columns)
+                    if t in validation_settings.keys()
+                ]
```
leewujung, line 450: This section gets the job done, but it is a little convoluted because you need to get the new column names but then have to fish out the corresponding old column names in order to pass those to `read_validated_data`.

Wouldn't it be more straightforward to combine this function and `read_validated_data` so that you only read the sheets once and can convert the column names directly as they are read in, and check if anything is missing there?

leewujung (approved):
I think this can be merged as is, but I feel that `validate_data_columns` and `read_validated_data` can be combined into a single function. I also feel like I've made this comment before?? or maybe I made the reverse of this comment and the functions were actually split out from one?? I don't remember...

echopop/core.py
leewujung: Oops, yes I highlighted the wrong lines! I did mean `Strata Index` and `Stratum`. Ok!

echopop/utils/load.py
```diff
@@ -386,79 +375,11 @@ def read_validated_data(
             input_dict["acoustics"]["nasc_df"][column_to_add] = df[column_to_add]
     else:
         raise ValueError(
-            """Unexpected data attribute structure. Check API settings located in"""
-            """the configuration YAML and core.py"""
+            "Unexpected data attribute structure. Check API settings located in "
+            "the configuration YAML and core.py."
```
leewujung, line 379: I think the configs are more like parameters, so minor suggestion to just leave out the word API.
```suggestion
            "Unexpected data attribute structure. Check the settings in "
            "the configuration YAML and core.py."
```

leewujung (approved):
Awesome! This looks a lot cleaner now. Other than my minor edit suggestion, feel free to merge!

PR #269, opened 2024-09-09
Update default `crop_method` value (for mesh)

brandynlucca:
This swaps the default argument for `crop_method` within the `Survey.kriging_analysis(..., crop_method)` method. The default was previously `"interpolation"`. This has been updated to `"transect_ends"`. This change has been reflected in the accompanying `example_echopop_workflow.ipynb` example notebook as well.


leewujung (approved):
LGTM! Thanks for getting this in!

Issue #270, opened 2024-09-10
inpfc stratum


rebeccathomas-NOAA:
I am having problems running survey.transect_analysis with a strata equals INPFC. I can get survey.transect_analysis() and survey.transect_analysis(stratum='ks') to run just fine.
However, when I run
survey.transect_analysis(stratum='inpfc'),
I get the error:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\rthomas\Pycode\echopop\echopop\survey.py", line 188, in transect_analysis
    self.analysis["transect"] = process_transect_data(
  File "C:\rthomas\Pycode\echopop\echopop\analysis.py", line 80, in process_transect_data
    aggregate_sigma_bs(
  File "C:\rthomas\Pycode\echopop\echopop\acoustics.py", line 161, in aggregate_sigma_bs
    aggregate_lengths = specimen_data.meld(
  File "C:\rthomas\Pycode\echopop\echopop\utils\operations.py", line 180, in meld
    specimen_dataframe.copy()
  File "C:\rthomas\Pycode\Learn_EchoPop\.venv\lib\site-packages\pandas\core\frame.py", line 9183, in groupby
    return DataFrameGroupBy(
  File "C:\rthomas\Pycode\Learn_EchoPop\.venv\lib\site-packages\pandas\core\groupby\groupby.py", line 1329, in __init__
    grouper, exclusions, obj = get_grouper(
  File "C:\rthomas\Pycode\Learn_EchoPop\.venv\lib\site-packages\pandas\core\groupby\grouper.py", line 1043, in get_grouper
    raise KeyError(gpr)
KeyError: 'inpfc'

I realize that this could be a user error, so I apologize if raising it as an issue is the wrong thing to do.
Thanks,
Rebecca

brandynlucca:
@rebeccathomas-NOAA: thanks for reporting this! I have labeled this as a priority item and will try and get it fully addressed today with a complete patch, so there should be a pull request specifically addressing this in the next day or so. I am adding a checklist below for my internal tracking. I'll reach out if I have any clarification Q's on this Issue.

- [ ] Isolate arguments/variables/dictionaries that incorrectly use `'inpfc'` in place of `'stratum_inpfc'`
- [ ] Test stratum-definition switching when toggling between `stratum = 'ks'` and `stratum = 'inpfc'` 
- [ ] Fix/replace/refactor/validate method for stratum-assignment for biological data
[...]

PR #272, opened 2024-09-11
Fix to re-enable `transect_analysis(..., stratum="inpfc")`

brandynlucca:
This PR addresses #270 reported by @rebeccathomas-NOAA where `Survey.transect_analysis(..., stratum="inpfc")` would fail at various points in the workflow due to conflicting columns (i.e. differentiating between `'stratum_num'` and `'stratum_inpfc'`). These issues have been patched to fix this bug in the computation when stratifying by INPFC instead of KS-clustering.


brandynlucca:
Remove the superfluous commented-out code.

leewujung (approved):

Issue #191, opened 2024-03-01
docs: Changes needed to "Input files"


leewujung:
- [x] typo: INPC --> INPFC
- [x] Empty value column --> can remove. Add a note on how missing values are treated (or what happens to those rows)
- [x] Units for stratum_num etc --> "index"
- [x] `fraction_hake` or similar variable: Units --> fraction, valid range: [0, 1]
- [x] Add an issue to remind ourselves add details for NASC folder batch processing once that functionality is added
- [x] Add a column "Location in data structure"
- [x] Modify the leading text under Biological (trawl) data to make the subfolder path explicit and say the structures are identical
- FEAT: ask RT to change `Latitude (upper limit)` to `northlimit_latitude` for the INPFC sheet in the "Geo-strata" xlsx file
- [x] Column name --> Echopop column name
- [x] Old column name --> EchoPro column name (based on 2019 files)
- [x] remove redundancy and what's already covered in other new doc pages

leewujung:
Moved this to v0.4.1 since many in the list have been addressed.

PR #274, opened 2024-09-14
Drop haul-transect/transect-region requirements

brandynlucca:
This PR applies a bug-fix to the NASC ingestion procedure by deprecating the creation of the haul-transect and transection-region key external files. This removes the need to intermittently generate these files since the haul-transect map is not required until the biological data are ingested. In theory, there may be a need to retain reading in the `biodata_gear.xlsx` file for conserving precise transect information for the biological data (i.e. for bootstrapping/randomization). But this is not necessary for the currently available workflow, and possibly avoids issues in the future for datasets where the `biodata_gear.xlsx` file(s) have `transect_num` missing entirely.


leewujung (commented):
Hey @brandynlucca : thanks for the PR, and sorry for the delay in review it. I put all my comments below instead of inline comments, since these touch upon broader context than the changes you made. Some are cosmetic (like reorganization) and we can discuss if it is necessary. I think the docs one will be good to have for v0.4.1 though.



## `read_validated_data`

This section below is identical in both the if and else conditions under `load_dataset` - can they be merged (moved outside of the if-else)?
```python
# Validate datatypes within dataset and make appropriate changes to dtypes
# ---- This first enforces the correct dtype for each imported column
# ---- This then assigns the imported data to the correct class attribute
read_validated_data(
    input_dict,
    configuration_dict,
    file_name,
    sheet_name,
    config_map,
    validation_settings,
)
```

## `map_imported_datasets`
`map_imported_datasets` is used first under `load_data` and then under `prepare_input_data`, but `prepare_input_data` is called in `load_data`. There seems to be some redundancy here?


## `prepare_input_data`

For `prepare_input_data` I have a few questions and thoughts:
1. Are the `if` conditions designed for the cases when only a subset of these exist? What are those conditions -- I was under the impression that all of these need to exist for the workflow to execute.
- Related to the above -- what happens when none of the `if` conditions is true? 
- In `prepare_input_data` there are clear sections handling different data types. The function is so long that I wonder if it makes sense to factor some of them out as standalone functions.
- Related to the above, since `prepare_input_data` is only used once under `load_data`, depending on what the answers of 1-2 are, it might make sense to make individual functions for each data type and just call all of them in `load_data`?

## docs
While reviewing the code, I was trying to understand what is in the config table and found terms such as "superlayer" and "name" pretty confusing. I couldn't find a documentation page that explains what's what in the config file -- is there one? If not, I think we should create such a page with a block diagram showing the flow of operations in `load_data`. There are multiple mapping and cleaning/subselection in the code, and this will help users/readers understand what is done and why, and potentially help with future refactoring/simplification.

## about loading NASC
I think this is the first time I skimmed through the NASC loading functions, so 2 questions:
- Are there currently tests for the result of `load_nasc` to compare values from those loaded into Matlab? If not, can that be done? Not sure if there is any bug in the Matlab code to prevent that.
- What is the difference between `read_echoview_export` and `batch_read_echoview_export`?

brandynlucca:
> This section below is identical in both the if and else conditions under load_dataset - can they be merged (moved outside of the if-else)?

I'm not sure what the best approach for accomplishing there is a nuanced difference in how the filenames and sheetnames are defined depending on whether `dataset == "biological"` where the `region_id` key is required versus `dataset != "biological"` where `region_id` is omitted entirely. 

_Action items_

- [ ] Default to converting all `sheet_name` entries to a `list` regardless of the dataset
- [ ] Consolidate how `file_name` is defined so it is done prior to the `if-else` and `for-loop` to reduce repetitiveness
- [ ] Reduce the scope of the `if-else` statement to only generating the `file_name` and `sheet_name`
- [ ] Refactor the related for loop (i.e. `for sheets in sheet_name`) to instead loop over `file_name-sheet_name` key pairs since  the actual validation of these files via the `Pydantic` models does not depend on whether the dataset is biological or something else. This will enable consolidating the code into a single `read_validate_data(...)` call (at least for each loop iteration)

> map_imported_datasets is used first under load_data and then under prepare_input_data, but prepare_input_data is called in load_data. There seems to be some redundancy here?

This function runs at the bookends of the data loading step leading up to the `prepare_input_data(...)`. So:

1. **[L126](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load.py#L126C21-L126C54)**: gathers current datasets already loaded into the `Survey` object (e.g. the case where `Survey.load_acoustic_data(...)` precedes `Survey.load_survey_data(...)` or vise versa). This is important for ensuring that only the defined input datasets are (re-)initialized without accidentally wiping other datasets (i.e. acoustic vs survey data).
2. **[L478](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load.py#L478)** [within `prepare_input_data(...)`]: gathers the updated set of datasets now loaded into the `Survey` object after `read_validate_data(...)` succeeds across all user-defined input files. This is important since the pre-processing step in `prepare_input_data(...)` can only run certain calculations when specific datasets are present. 

The outputs at each step are not necessarily identical; however, I think this can be refactored to avoid the duplication. Some of this relates to some of your later comments/suggestions.

_Action items_
- [ ] Append successfully imported datasets to the `imported_data` list generated on **[L126](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load.py#L126)** at the conclusion of the **[dataset loading loop](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load.py#L174-L251)**
- [ ] Add the output of `map_imported_datasets` as an argument for `prepare_input_data`. The alternative would be to filter out empty dictionary keys associated with a particular dataset within `prepare_input_data`, but that seems more convoluted than enabling a simple `list` input

>Are the if conditions designed for the cases when only a subset of these exist? What are those conditions -- I was under the impression that all of these need to exist for the workflow to execute.

The `if` conditions are designed for cases when only a subset of the datasets exist. This allows for specific file inputs to be specified (e.g. just `biology` versus `biology`, `spatial`, and `statistics` in a single step). Moreover, it is necessary for decoupling calculations that require only the inputs from `Survey.load_acoustic_data(...)`, those from `Survey.load_survey_data(...)`, and finally those that require both `Survey.load_acoustic_data(...)` and `Survey.load_survey_data(...)`. This may have been the result of some other facet of the code, but these conditional statements also solved an issue where running `Survey.load_survey_data(...)`/`Survey.load_acoustic_data(...)` a second time raises various Errors messages, erroneously concatenates duplicate dataframes, etc. This is similar for the case where different issues could arise if one runs `Survey.load_acoustic_data(...)` or `Survey.load_survey_data(...)` first. 

> Related to the above -- what happens when none of the if conditions is true?

In theory, that shouldn't be possible since the [ `CONFIG_DATA_MODEL`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/validate_dict.py#L448) ensures that the filesets exist in the configuration and [`DATASET_DF_MODEL`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/validate_df.py#L487) ensures that the loaded data files comport with formats required by `Echopop`. So if both of those validation steps pass all checks, then at least one of the conditions within `prepare_input_data` should be satisfied since `imported_data` should not comprise an empty `list`. 

But in an edge case where `imported_data = []`, then that would indicate that no data were loaded (or I suppose that the user deliberately deleted all of the data from the `Survey` object) and would raise an Error when trying to then run `Survey.transect_analysis(...)`. In this case, I suppose it would make sense to put together an integrity checker either at the end of `Survey.load_{survey/acoustic}_data(...)` to verify that all datasets exist _or_ include the checks at the start of `Survey.transect_analysis`, `Survey.fit_variogram`, `Survey.kriging_analysis`, and `Survey.stratified_analysis`. 

_Action items_
- [ ] Write an integrity checker that validates successful loading (or at least presence) of datasets that are required for the various analysis methods for the `Survey` object

>In prepare_input_data there are clear sections handling different data types. The function is so long that I wonder if it makes sense to factor some of them out as standalone functions.

This was originally included as one of the orchestrated workflows in `analysis.py` but was then later moved to `utils.load.py`. It makes sense to refactor the various parts of `prepare_input_data` into separate functions for legibility purposes. I think it would still be best to keep all of these wrapped within `prepare_input_data` since that would otherwise require including the conditional statements into the broader `load_dataset` function.

_Action items_
- [ ] Refactor each conditional clause within `prepare_input_data` into separate functions
- [ ] Replace the function name `"prepare_input_data"` with something else like `"preprocess_inputs"` or something to that effect

leewujung:
> > While reviewing the code, I was trying to understand what is in the config table and found terms such as "superlayer" and "name" pretty confusing. I couldn't find a documentation page that explains what's what in the config file -- is there one? If not, I think we should create such a page with a block diagram showing the flow of operations in load_data. There are multiple mapping and cleaning/subselection in the code, and this will help users/readers understand what is done and why, and potentially help with future refactoring/simplification.
> 
> Ah yes, this corresponds to the reference dictionary **[`LAYER_NAME_MAP`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/core.py#L162)** that is used for renaming particular keys. This was done to reconcile the preexisting fileset keynames contained within `survey_year_{YEAR}_config.yml` and the actual `Survey` keys within the `Survey.input` attribute. But I agree that including related documentation for this would help clarify things. This also relates to #191 since part of this includes renaming column names, etc.

I think we should consider in the not very distant future, reconcile the difference between `survey_year_{YEAR}_config.yml` and the `Survey` object structure, so that the implementation becomes more straightforward. Especially since these are all under our control (i.e. there is no reason we cannot change the yaml file). Another thing I would recommend is to flatten the nested structures, either directly make the yaml more flat (so different types of data have the same number of depth), or in the code right after the yaml is read into the `Survey` object (so each attribute/dictionary for different data types has the same number of depth).

### 
> > What is the difference between read_echoview_export and batch_read_echoview_export?
> 
> [`read_echoview_export`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load_nasc.py#L16-L100) is used for reading the literal `*.csv` files and applying a rolling imputation to bad coordinates (as done in `EchoPro`). This function is then contained within the [`consolidate_exports`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load_nasc.py#L177-L234) function (i.e. **[L189](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load_nasc.py#L189)**) for consolidating the all of the `*.csv` files for a given export filetype (e.g. `analysis`, `cells`). [`batch_read_echoview_exports`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load_nasc.py#L422-L578) is the primary function called by `Survey.load_acoustic_data(ingest_exports="echoview")` that first consolidates all of the exports (i.e. [`consolidate_exports`](https://github.com/brandynlucca/echopop/blob/4aceb41cce2b8e8306334818bbc1c76bc7136a63/echopop/utils/load_nasc.py#L482-L484)), merges the transect and haul information into the consolidated export data, and fills a template interval DataFrame (i.e. all k intervals) with the consolidated exports that only represent where hake were present.

Ok, I see. How about changing `batch_read_echoview_exports` to just `ingest_echoview` to avoid confusion?


###
> > Are there currently tests for the result of load_nasc to compare values from those loaded into Matlab? If not, can that be done? Not sure if there is any bug in the Matlab code to prevent that.
> 
> I've confirmed that the data structures and values are nearly identical between `Echopop` and `EchoPro` (at least for the 2019 dataset on both my end and for the files generated by RT on her end).

Ok, let's go over the sources of differences when we talk.

brandynlucca:
> Ok, I see. How about changing batch_read_echoview_exports to just ingest_echoview to avoid confusion?

I committed the name change from `batch_read_echoview_exports` to `ingest_echoview_exports`. 

> I think we should consider in the not very distant future, reconcile the difference between survey_year_{YEAR}_config.yml and the Survey object structure, so that the implementation becomes more straightforward. Especially since these are all under our control (i.e. there is no reason we cannot change the yaml file). Another thing I would recommend is to flatten the nested structures, either directly make the yaml more flat (so different types of data have the same number of depth), or in the code right after the yaml is read into the Survey object (so each attribute/dictionary for different data types has the same number of depth).

This has been added as a tracking issue in #282.

PR #292, opened 2024-10-29
Add references page to sidebar

leewujung:
This adds the references page to the side bar. It was previously rendered by not linked explicitly, and was only accessible via the stratification page.


leewujung:
Fixed the wrong page name, I'll self-merge this now!

PR #293, opened 2024-11-02
Update `setup.py` for automatic versioning

leewujung:
Update `setup.py` to fix versioning problem. Goal is to make it be based on git tags.


leewujung:
@brandynlucca - not sure if you're "watching" this repo so just pinging you here as an FYI.

PR #295, opened 2024-11-04
Add plotting methods

brandynlucca:
This PR includes:

* A new `plot` method for `Survey`-class objects
* Several plotting options for population estimates
  - Bivariate age-length distribution 
  - Along-transect distributions
  - Kriged (mesh) population estimates
* Additional data exploration/manipulation options specifically for the kriged mesh estimates
  - Additional `plot_type` options with the default being `plot_type="hexbin"`. This enables different views into the dataset such as being able to display the standard deviations of binned data, etc. 
* Associated `Pydantic` validators for `echopop`-specific parameters
* `numpydocs`-formatted docstring with additional documentation and details
* An accompanying `jupyter-book` page that provides an example notebook with different options/functionalities
  - This has been added to the `_toc.yml` table of contents configuration and successfully renders locally


echopop/graphics/plotting.py
```diff
@@ -0,0 +1,769 @@
...
+def plot_age_length_distribution(
+    data_dict: Dict[str, Any],
+    **kwargs,
+):
+
+    # Get the dataset variable name
+    variable = kwargs.get("variable", None)
+
+    # Get sex
+    sex = kwargs.get("sex", None)
```
leewujung, line 641: Why do you need these if the inputs are already validated? If either of the variables are `None`, the subsequence code would error out also.

echopop/graphics/plotting.py
```diff
@@ -0,0 +1,769 @@
...
+    kwargs.update(dict(xlabel=kwargs.get("xlabel", "Age (years)")))
+    # ---- y
+    kwargs.update(dict(ylabel=kwargs.get("ylabel", "Fork length (cm)")))
+
+    # Initialize figure
+    # ---- Update the 'figsize' if it doesn't yet exist
+    if not kwargs.get("figsize", None):
+        kwargs.update(dict(figsize=kwargs.get("figsize", apply_aspect_ratio(5.5, axis_limits))))
+    # Initialize figure
+    fig, ax = plt.subplots(**prune_args(plt.subplots, **kwargs))
```
leewujung, line 701: What is the purpose of `prune_args`? Seems like it is to select only the parameters that subplot can accept?

echopop/graphics/plotting.py
```diff
@@ -0,0 +1,769 @@
...
+    elif "nasc" in variable:
+        cmap = kwargs.get("cmap", "cividis")
+        colorbar_label = kwargs.get("colorbar_label", "NASC\n$\\mathregular{m^{2}~nmi^{-2}}$")
+    # ---- Number density
+    elif "number_density" in variable:
+        cmap = kwargs.get("cmap", "magma")
+        colorbar_label = kwargs.get(
+            "colorbar_label",
+            (label_prepend + "number density\n$\\mathregular{animals~nmi^{-2}}$").capitalize(),
+        )
```
leewujung, line 295: Since the variables are the same, seems like you can use a dict structure to store the default values for different variables and use the same `kwargs.get()` structure to substitute user-supplied values.

Same comments for the code under `plot_mesh`.

echopop/survey.py
```diff
@@ -658,6 +658,148 @@ def kriging_analysis(
...
+        # Proceed with plotting
+        # ---- Type: spatial
+        if plot_info["type"] == "spatial":
+            # ---- Get the geospatial configuration
+            geo_config = self.config["geospatial"].copy()
+            # ---- Get the coastline, if it exists, from `plot_parameters`
+            if parameters.get("geo_config"):
+                geo_config.update(parameters.get("geo_config", None))
+            # ---- Update the parameterization
+            parameters.update(dict(geo_config=geo_config))
```
leewujung, line 792: A few comments:
- do you need to define a new variable `plot_parameters`? 
- there is a comment `"Get the coastline, if it exists, from plot_parameters"`, but the operation would update the `parameter` dict with _everything_ in `geo_config`, not just coastline
- the if check for `parameters.get("geo_config")` is confusing: when `parameters["geo_config"]` does not exist, the statement `geo_config.update(parameters.get("geo_config", None))` will not be executed, but in there you have in there to update the dict `geo_config` with None. I am not following what you are trying to achieve here, but I think there's some redundancy here that can be removed.

leewujung (commented):
@brandynlucca : Thanks for the PR! I made some inline comments above.

Overall I think this PR is ready to go. I have 2 main comments:

-------------------
I think it would be good to actually expose functions like `plot_mesh`, `plot_transect`, `plot_age_length_distribution` in the `Survey` object, in a way similar to how you can access the plot functions in pandas. This is something that I think should be tabled for now so that we can get other parts of the codebased built. I will create an issue for it as a reminder -- I think one can use accessor functions to streamline the process. 

-------------------
I feel the way the allowed variables are restricted for each plot type is a little limiting. For example, why isn't kriged biomass density in the list of the mesh format? but that is "kriged_mean", right? and you also hard-coded `kriged_variable = "biomass"` in the `plot_mesh` function, so there's a potential confusion on the user side of what is being plotted.

My view of the plotting functions is that they are like "containers" where the variables can be swapped out as long as the dimension of the variables fit what is needed for the plot to happen, like how matplotlib `imshow` or `scatter` works. Plotting any variables with corresponding lat/lon values in the mesh format should be the same as plotting one of the variations in the defined set of allowed variables.  I can see the convenience of providing defaults for the ticks and labels, which is great. I just think that leaving the door open for other variables would enhance reusability of the functions.

Using the kriged biomass density as an example, I feel it would be useful if the users can provide a 3-column dataframe (value, lat, lon) and specify that it is kriged biomass density, and can get the automatic ticks and labels. If they do not specify the type, they don't get any automatic labels.

Let me know what you think.

echopop/graphics/plotting.py
brandynlucca: This is required because otherwise any and all `kwargs` supplied by the users error out for the various `matplotlib` functions. For instance, including `"figsize"`, `"vmin"`, and `"vmax"` in `kwargs` raises an Error when `kwargs` is passed to `matplotlib.pyplot.subplots` due to the inclusion of `"vmin"` and `"vmax"` despite `"figsize"` being a valid keyword argument. So `prune_args` subsets the parameters within `kwargs` that are a valid subset of those for the various functions where `prune_args` is used. I already tried to be as detailed as possible in the documentation for the `Survey.plot` method concerning these keyword args. I'll have to mull this over a bit more to think about alternative parameterizations.

echopop/survey.py
brandynlucca: > do you need to define a new variable plot_parameters?

This was done to keep things a bit tidier and to streamline how plotting parameters are fed into the `PlotModel` Pydantic validator. One alternative would be to keep `plot_parameters` for `Echopop`-specific plotting parameters and allow for `**kwargs` to include whichever `matplotlib` keyword arguments.

> there is a comment "Get the coastline, if it exists, from plot_parameters", but the operation would update the parameter dict with everything in geo_config, not just coastline

The purpose of this chunk is to pull in geospatial parameterization stored within the `Survey` object (primarily the projection, which in this case provides a consistent usage of the same projection between spatial calculations and plotting).

```python
from echopop.survey import Survey

# Mock
self = Survey
self.config = {"geospatial": {"init": "epsg:4326"}}
parameters = {"vmin": 0, "vmax": 100, "geo_config": {}}
```

This yields: `{'vmin': 0, 'vmax': 100, 'geo_config': {'init': 'epsg:4326'}}`

```python

# Mock
self = Survey
self.config = {"geospatial": {"init": "epsg:4326"}}
parameters = {"vmin": 0, "vmax": 100, "geo_config": {"init": "epsg:9999"}}
```

This yields: `{'vmin': 0, 'vmax': 100, 'geo_config': {'init': 'epsg:9999'}}`

```python

# Mock
self = Survey
self.config = {"geospatial": {"init": "epsg:4326"}}
parameters = {"vmin": 0, "vmax": 100}
```

This yields: `{'vmin': 0, 'vmax': 100, 'geo_config': {'init': 'epsg:4326'}}`

So this is the intended interaction, although I think it may be redundant since the `PlotModel` `pydantic` model will provide a default "EPSG" projection value if none is supplied. This therefore may be unnecessary. 

> the if check for parameters.get("geo_config") is confusing: when parameters["geo_config"] does not exist, the statement geo_config.update(parameters.get("geo_config", None)) will not be executed, but in there you have in there to update the dict geo_config with None. I am not following what you are trying to achieve here, but I think there's some redundancy here that can be removed.

This is related to my last point above re: this check potentially being redundant and therefore unnecessary. The logic is:

1. Did the user provide a geo-configuration in `plot_parameters`? If so, update the stored geo-configuration from the `Survey` object. 
2. Regardless of the presence for `geo_config` in `plot_parameters`, whatever geo-configuration stored within the `Survey` object and those supplied in `plot_parameters` are combined. 

With that in mind, that is a goofy way to approach what I was trying to accomplish here, so I agree that there is some circular dog-chasing-tail behavior going on here that can be clipped.

brandynlucca:
Adding tracking checkboxes:

- [ ] Reference dictionary/etc. for variable `kwargs` (e.g. `cmap`, `colorbar_label`, etc.)
- [ ] Remove redundancy in geo-configuration plotting parameters check
- [ ] Give kriged mesh variable plotting options clearer names
- [ ] Address `kriged_variable` and `units` variables in `plot_mesh`
- [ ] Update example workbook
- [ ] Update documentation

echopop/graphics/plotting.py
leewujung: Yes, I know that this is to avoid users erroring out, but I don't think it's a good idea that the users actually put in erroneous arguments and do not know, and they would be left wondering why the command does not work.

I think the easiest way to solve this is to print a warning on anything that has been pruned because it is not allowed in matplotlib, and they will have no effects.

brandynlucca: Agreed. I'll add that to the checklist on things to incorporate!

PR #301, opened 2024-11-16
Report generation

brandynlucca:
This PR adds: 

* New reports generation module (`echopop.extensions.FEATReports`) which comprises a series of helper functions that all feed into a central `FEATReports` class
* The associated `*.xlsx` files are produced at the defined directory path via `FEATReports.generate()`, which can be tuned to produce a subset of the reports if needed
* Fleshed out docstring documentation
* An example notebook (and therefore documentation) to demonstrate how `FEATReports` is used for generating these reports


docs/_toc.yml
```diff
@@ -30,6 +30,7 @@ parts:
     sections:
     - file: implementation/echopro_vs_echopop
     - file: implementation/preprocessing_data
+    - file: implementation/report_generation.ipynb
```
leewujung, line 33: Nothing wrong with this, but you can do without `.ipynb`

echopop/extensions/FEATreports.py
```diff
@@ -0,0 +1,1186 @@
...
+    reports: List[str]
+        A list of the report-types that will be produced
+
+    save_directory: pathlib.Path
+        A Path object comprising the save directory path where report files will be written
+
+    Methods
+    -------
+    *generate()*
+        Report factory method thaqt generates the `*.xlsx` files used by the FEAT team
```
leewujung, line 567:
```suggestion
        Report factory method that generates the `*.xlsx` files used by the FEAT team
```

brandynlucca:
Creating a tracking checklist:

- [ ] In-line changes
- [ ] Report output folder
- [ ] Report generation API
- [ ] __METHOD_REFERENCE__

PR #304, opened 2024-11-20
Diagnostic plotting extension

brandynlucca:
This PR adds:

* A new diagnostic plotting module (`echopop.extensions.diagnostics`) which includes a central `DiagnosticPlot` class
* An example notebook to demonstrate the different available diagnostic plots


echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+                        "longitude"
+                        if "longitude" in col
+                        else "latitude" if "latitude" in col else col
+                    )
+                    for col in v.columns
+                },
+                inplace=True,
+            )
+            for k, v in DATAINPUT["Latitude/longitude"].items()
+        ]
```
leewujung, line 128: Since this file `self.data.input["statistics"]["kriging"]["isobath_200m_df"]` is part of Echopop standard input, I think it's better to force the excel file/datafram to be of a specific format instead of doing this type of automatic 
 and "silent" reformatting. I am not a big proponent of silent reformatting in anticipation of user errors -- I think people need to know what they put into the analysis.

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+                "xlab": "Normalized x",
+                "ylab": "Normalized y",
+            },
+        }
+
+        # Initialize `panel` extension
+        pn.extension()
+
+        # Create dropdown selector
+        var_dropdown = pn.widgets.Select(name="Variable", options=var_options)
```
leewujung, line 159: Seems better named as "Coordinate"
```suggestion
        # Create dropdown selector
        var_dropdown = pn.widgets.Select(name="Coordinate", options=var_options)
```

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+
+            # Initialize plotting layers
+            plot_layers = []
+
+            # Plot the data
+            # ---- Base tile layer
+            if tile:
+                plot_layers.append(tile)
+            # ---- Full mesh
+            if "full" in data:
```
leewujung, line 184: Just adding this for readability
```suggestion
            if "full" in data:  # not shown for isobath-referenced coord
```

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+        plot_stacked = (self.tile * HOVERPATH * gv.Overlay(plot_layers)).opts(
+            width=800, height=800, xlabel="Longitude (\u00B0E)", ylabel="Latitude (\u00B0N)"
+        )
+        # ---- Return
+        return plot_stacked
+
+    def nasc_map(self) -> hv.core.overlay.Overlay:
+
+        # Get the full dataset
+        DATAINPUT = self.data.analysis["transect"]["acoustics"]["adult_transect_df"].copy()
```
leewujung, line 277: I wonder if it is useful to make `"adult_transect_df"` the default but allow plotting other df names -- I forgot if there are other dfs in `self.data.analysis["transect"]["acoustics"]`.

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+                    "transect": pd.DataFrame(
+                        {
+                            "stratum": strata_info.stratum_inpfc,
+                            "density": strat_krig["estimate"]["strata"]["density"] * 1e-6,
+                            "lower": [k[0] * 1e-6 for k in strat_krig["ci"]["strata"]["density"]],
+                            "upper": [k[1] * 1e-6 for k in strat_krig["ci"]["strata"]["density"]],
+                        }
+                    )
+                }
+            )
```
leewujung, line 420: You can use a for loop to set both "kriging" and "transect" entries:
```
for agg_type in ["kriging", "transect"]:
  if agg_type in self.data.results:
   ...
```
Since the majority of the code is identical, this will make it more robust to potential typos.

leewujung: Along the same line, if you use the same name for the quantity, you don't need to separately handle `var_columns = {"Biomass": "total", "Biomass density": "density"}` in the plotting

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def stratified_results(self) -> pn.Column:
+
+        # Get the full datasets
+        # ---- Get stratification information
+        strata_info = self.data.input["spatial"]["inpfc_strata_df"].copy()
+        # ---- Initialize
+        DATAINPUT = {"Biomass": {}, "Biomass density": {}}
+        # ---- Kriging
+        if "kriging" in self.data.results:
```
leewujung, line 364: Do you anticipate cases when "kriging" or "transect" is NOT in `self.data.results`?

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+                    ).opts(
+                        size=12,
+                        color="black",
+                    )
+                    # ---- Add to layer
+                    plot_layers.append(transect_pt_err * transect_pts)
+                # ---- Layer the data
+                return hv.Overlay(plot_layers).opts(
+                    height=600, width=800, xlabel="INPFC stratum", ylabel=y_axis_label[selection]
+                )
```
leewujung, line 490: These are indented -- are the indents intentional? If so, then transect data will only be plotted if kriging analysis has been run - this would contradicts with the multiple `if "kriging" in self.data.results:` and `if "transect" in self.data.results:` checks you did earlier ([see my comments here](https://github.com/OSOceanAcoustics/echopop/pull/304/files?diff=unified&w=0#r1894709099)).

echopop/extensions/diagnostics.py
```diff
@@ -0,0 +1,510 @@
...
+            # ---- Kriging
+            if "kriging" in data:
+                # ---- Add errorbar
+                kriging_pt_err = hv.Segments(
+                    data["kriging"], kdims=["stratum", "lower", "stratum", "upper"], label="Kriging"
+                ).opts(color="red")
+                # ---- Point estimates
+                kriging_pts = hv.Points(
+                    data["kriging"], kdims=["stratum", value], label="Kriging"
+                ).opts(
```
leewujung, line 460: I suggest slightly offsetting the black and red points horizontally so that error bars can be easily seen and compared. It is hard to see right now with the overlap, like what's shown in the example fig.

leewujung (changes requested):
@brandynlucca : Thanks for the PR. Sorry for the lag in reviewing this.

I think the code using hvplot functionalities looks great in general, and the notebook illustrates the diagnostic plots very well - the pre-generated figures are very helpful. I have a few inline comments and a couple other points below.

Once these are addressed feel free to go ahead and release v0.4.2!

## Kriging function names
I feel I have made the same/similar comments some time earlier, but I don't remember where - it may also just be when I was forcing the ML prediction data through the mesh cropping and kriging. 

There are many functions with very similar names, and it would be good to make them less confusing. If I read these correctly, there functions break the kriging operations into multiple steps, but the name does not reflect that:
- `analysis.py::krige`: clean up and update settings to use for kriging (mainly `settings_dict`)
- `krige.py::kriging`: set up spatial grids and associated parameters
- `krige.py::kriging_interpolation`: actually performs the kriging operation

Perhaps change the names and adding more descriptive text in the docstring so that it is clearer what each function does.


## Paths in diagnostic plot notebook
```
from echopop.survey import Survey

init_config = "C:/Users/Brandyn/Documents/GitHub/echopop/config_files/initialization_config.yml"
file_config = "C:/Users/Brandyn/Documents/GitHub/echopop/config_files/survey_year_2019_config.yml"
```
Changing all cases like these to using relative path to a folder under the repo would be nice.

echopop/extensions/diagnostics.py
brandynlucca: There are a few other dataframes that get added to the `"acoustics"` dictionary (i.e. `sigma_bs` averaged over hauls and strata), but those are more tabular and only provide the mean values (I think it would only be useful to have a dataviz for these if there were standard errors or some other uncertainty interval associated with the values).

echopop/extensions/diagnostics.py
brandynlucca: Good point. For this function to run, I think it would only be reasonable to anticipate `"kriging"` to be missing since it inherits values from `"transect"`. I can add a warning message here that states as much.

echopop/extensions/diagnostics.py
brandynlucca: > Along the same line, if you use the same name for the quantity, you don't need to separately handle var_columns = {"Biomass": "total", "Biomass density": "density"} in the plotting below, as that's the only difference between the biomass and biomass density plots.

I'm not sure I fully follow this comment/suggestion. The `"total"` and `"density"` results are stored in separate dictionaries under both `"transect"` and `"kriging"`, so I'm not sure how using the same name would be viable here?

brandynlucca: Oh, unless you are referring to how the column for the values for the DataFrame within `DATAINPUT` is named. In that case, I totally agree that using a shared column name like `"mean"` would make more sense.

leewujung: Yes, column names were what I meant, since then it wouldn't require separate handling.

PR #321, opened 2025-01-10
Kril inversion component sketch

leewujung:
Creating this so that it is easier to comment on code/pseudocode.


echopop/extensions/inversion/optimize.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+from lmfit import Minimizer, Parameters
+from typing import Any, Callable, Dict, Literal, Union
+from numpy.typing import ArrayLike
+
+def mae(
```
leewujung, line 6: should this be `mad`? it's used below in `prepare_optimization`

leewujung, line 8: should this be `mad`? it's used below in `prepare_optimization`

echopop/extensions/inversion/optimize.py
```diff
@@ -0,0 +1,75 @@
...
+    # PHASE 1) EXTRACT RELEVANT SCATTERING MODEL PARAMETERS
+    # PHASE 2) CONVERT RELEVANT OPTIMIZATION PARAMETERS INTO ASSOCIATED `lmfit::Parameters`
+    params = Parameters(**scattering_model_parameters) # not actual code, just a placeholder
+    # PHASE 3) WITH COST-FUNCTION, CREATE `lmfit::Minimizer` OBJECT
+    # not actual code, just a placeholder 
+    minim = Minimizer(cost_function, params, **optimization_settings)         
+    # RETURNS: Dictionary with optimization parameters and minimizer
+    return {"parameters": params, "minimizer": minim}
+
+def optimize_scattering_model(
```
leewujung, line 56: This should be renamed to something like `optimize_model_parameters`, since it is not the model that gets optimized, but the parameters.

echopop/extensions/inversion/scattering_models.py
```diff
@@ -0,0 +1,52 @@
...
+    # == Scat_models/DWBAbscat1.m
+
+    # PHASE 1) EXTRACT AND PREPARE PARAMETERS
+    # PHASE 2) COMPUTE CALCULATIONS
+
+    # RETURNS: np.ndarray of complex form function values
+    pass
+
+
+class validate_pcdwba(BaseModel):
```
leewujung, line 46: I think it'll be more efficient to have build a validator that works with a model class - let's talk more about this when we meet, because I had some thought on existing scattering model packages and had a discussion on this with Sven some time in Dec.

echopop/extensions/inversion/invert.py
```diff
@@ -0,0 +1,47 @@
+from typing import Any, Callable, Dict, Literal, Union
+
+import pandas as pd
+from numpy.typing import ArrayLike
+
+
+def normalize_scattering_model_parameters(
```
leewujung, line 7: I think this can go into `optimize.py` since the output of the optimizer is doing the "inversion" so can just be packaged there.

echopop/extensions/inversion/invert.py
```diff
@@ -0,0 +1,47 @@
...
+    measured_Sv: ArrayLike[float],
+    predicted_Sv: ArrayLike[float],
+):
+    """
+    Compute inverted volumetric backscattering strength ($S[v]$) prediction error
+    """
+    pass
+
+
+def invert_population(
```
leewujung, line 27: This `invert_population`  and `Sv_prediction_error` are not strictly part of the "inversion" but are derived computations. Let me read through the rest of doc and see where these might go.

echopop/extensions/inversion/patcher.py
```diff
@@ -0,0 +1,48 @@
+from pathlib import Path
+from typing import Any, Dict, Union
+
+import pandas as pd
+
+from ...survey import Survey
+
+
+def inversion_pipeline(
```
leewujung, line 9: The way things are designed now are _completely separate_ from the hake component, which is not what I envisioned. I think it is fine as an intermediate step, but ultimately a more modularized framework will be needed to make the package more useful as an open-source project. Let's talk more about this when we meet.

leewujung:
@brandynlucca : Just saw this being closed. Can you create a new PR with the new design so that I see the revised designs re my comments, and the simplified scattering model and inversion functions? Thanks.

PR #349, opened 2025-05-08
NASC ingestion refactoring

brandynlucca:
This draft PR includes refactored changes to the NASC ingestion, docstrings, and associated tests.


echopop/nwfsc_feat/ingest_nasc.py
```diff
@@ -525,22 +525,53 @@ def merge_echoview_nasc(
...
+    """
 
-# def read_transect_region_file() -> pd.DataFrame:
-#     """
-#     Read external transect region definition file (if provided).
+    # Determine appropriate file reader
+    if filename.suffix == ".csv":
+        transect_region_df = read_csv_file(filename)
+    else:
+        transect_region_df = read_xlsx_file(filename, sheetname)
```
leewujung, line 565: I'd suggest just removing these and using `pd.read_csv` and `pd.read_excel` directly since the other line in both functions `export_file.columns = export_file.columns.str.lower()` is identical for both functions and can be easily added here.

leewujung:
@brandynlucca : testing structure looks great! thanks!

leewujung:
Oh also just noticed the merge conflict - seems just small things from my PRs added after your branched out this, so they ended up not in the commit history here.

brandynlucca:
> Why are setting column names to lowercase and impute_bad_coordinates needed in read_nasc_file? I was thinking df_nasc_all_ages (or df_nasc_no_age1) would be a "final" product so that people could just load and use those directly. Are the formatting and data in df_nasc_all_ages kept in a specific way to accommodate something else?

This mostly ensures backwards compatibility with previous FEAT survey years where the column name schemes are somewhat inconsistent. This is somewhat in anticipation of incorporating the validation step(s), which would incorporate any required formatting changes. 

> Is filter_transect_intervals related to the mesh polygon creation downstream? Or is it just so that people have some control over what part of the geographical regions are included?

This relates to removing off-effort transect intervals.

PR #350, opened 2025-06-05
Refactor survey data loading

brandynlucca:
This includes functions (with `pytest` tests) used for loading biological data, stratification files, and geostatistics parameters. I also updated `feat_hake.py` to reflect the updated usage for the various functions.

### DIFF
--- a/echopop/workflow/feat_hake.py
+++ b/echopop/workflow/feat_hake.py
@@ -18,10 +18,8 @@ filename_transect_pattern: str = r"T(\d+)"
 default_transect_spacing: float = 10.0  # nmi
 default_transect_spacing_latitude: float = 60.0  # deg N
 
-# Outputs:
-# ---- 1. Complete intervals regardless of hake region presence/absence [interval_df]
-# ---- 2. DataFrame containing the full set of cells-intervals-layers [exports_df]
-interval_df, exports_df = ingest_nasc.merge_echoview_nasc(
+# Merge exports
+df_intervals, df_exports = ingest_nasc.merge_echoview_nasc(
     nasc_path,
     filename_transect_pattern,
     default_transect_spacing,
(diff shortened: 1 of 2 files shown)


echopop/workflow/feat_hake.py
```diff
@@ -163,36 +164,177 @@
...
+# -------- INPFC (from strata)
+nasc_all_ages_df_inpfc = load_data.join_strata_by_haul(nasc_all_ages_df, 
+                                                       dict_df_strata["inpfc"])
+dict_df_bio_inpfc = load_data.join_strata_by_haul(dict_df_bio,
+                                                  dict_df_strata["inpfc"])
+# -------- KS (from strata)
+nasc_all_ages_df_ks = load_data.join_strata_by_haul(nasc_all_ages_df, 
+                                                    dict_df_strata["ks"])
+dict_df_bio_ks = load_data.join_strata_by_haul(dict_df_bio,
+                                               dict_df_strata["ks"])
```
leewujung, line 276: Why can these just be with the same name `df_nasc_all_ages` (note I have moved df to the beginning - following the pattern that the modifier goes to the end) but with different columns having the INPFC or KS stratum assignment?

leewujung (changes requested):
@brandynlucca : I made a couple suggestions that should be quick to address. Please resolve this first thing on Monday. Thanks.

echopop/workflow/feat_hake.py
brandynlucca: I suppose that they could. This was, in large part, just added to demonstrate the different uses to get to each of the stratum definitions and apply them to the underlying dataset. The reason I kept the two strata separate was primarily to let downstream functions reference the column `stratum_num`. I suppose I could add an additional argument that allows for naming the new stratum so multiple strata definitions could be applied to a preexisting dataframe, e.g.,

```python
load_data.join_strata_by_haul(df_nasc_all_ages, df_dict_strata["ks"], stratum_name="stratum_ks")
load_data.join_strata_by_haul(df_nasc_all_ages, df_dict_strata["inpfc"], stratum_name="stratum_inpfc")
```

And then downstream functions that incorporate stratum into their calculations would just need to be supplied the stratum column name?

leewujung:
Yes I think it will be cleaner to have both the INPFC and KS stratum information in the same dataframe, and just use the stratum name to select which one to use. This way users can easily inspect a single dataframe and see what the different stratum assignment looks like and don't have to keep track of many dataframes and dictionaries. 

I'll add an issue so that it gets into the backlog as part of refactoring .

PR #355, opened 2025-07-03
Initial apportionment calculations 

brandynlucca:
This includes functions, tests, etc. for accounting for age-1 (and other data subsets) contributions toward survey NASC, abundance, and biomass. This also sets up the additional apportionment required for the report generation.


brandynlucca:
@leewujung I originally said that this one could be skipped over, but I think it actually works just fine here. What this PR does _not_ include is applying the proportions to the transect data. That is an added component in #370.

echopop/nwfsc_feat/biology.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def set_population_metrics(
+    df_nasc: pd.DataFrame,
+    metrics: List[str] = ["abundance", "biomass", "biomass_density"],
+    stratify_by: str = "stratum_num",
+    df_average_weight: Optional[Union[pd.DataFrame, float]] = None,
+) -> None:
+    """
+    Convert acoustically derived number densities into population metrics (in-place)
```
leewujung, line 193:
```suggestion
    Calculate population metrics based on acoustically derived number densities.
```

echopop/nwfsc_feat/biology.py
```diff
@@ -0,0 +1,304 @@
...
+        # Ensure weights are properly aligned with the NASC data
+        if isinstance(df_average_weight, (pd.DataFrame, pd.Series)):
+            df_average_weight = df_average_weight.copy()
+
+            # If weights have the stratify_by column but not as index, set it
+            if hasattr(df_average_weight, "columns") and stratify_by in df_average_weight.columns:
+                df_average_weight.set_index(stratify_by, inplace=True)
+
+            # Reindex to match NASC data
+            df_average_weight = df_average_weight.reindex(df_nasc.index)
```
leewujung, line 282: I think the checking for `df_nasc` and checking for `df_average_weight` should be under separate if-else blocks.
Also the current if-else doesn't handle what happens when `stratify_by` does not exist in `df_nasc`.

For `df_average_weight`: It is kinda forced that you are checking if it has a column with the string in `stratify_by` and try to correct it here. I suggest going back to the upstream function (`stratum_averaged_weight`) to make sure the output is always of the same form. 

What are the other possibilities of the `stratify_by` string other than `stratum_num`? Since it's generated by `stratum_averaged_weight`, this string is fully under your control, so if there are multiple strings, it should be limited choices. In that case, the first check would naturally be if the `stratify_by` string is in the allowable set.

echopop/nwfsc_feat/biology.py
```diff
@@ -0,0 +1,304 @@
...
+    else:
+        if isinstance(result, pd.Series):
+            return result.to_frame("all")
+        else:
+            return result
+
+
+def set_population_metrics(
+    df_nasc: pd.DataFrame,
+    metrics: List[str] = ["abundance", "biomass", "biomass_density"],
```
leewujung, line 188: I suggest to remove these options and just calculate everything. It's not a lot of computation and in most cases people would just calculate all of them. This is especially true since you need abundance to get to biomass.

echopop/nwfsc_feat/biology.py
```diff
@@ -0,0 +1,304 @@
...
+        pivot_result = result.pivot(index="length_bin", columns=group_cols, values="weight_fitted")
+        return pivot_result
+    else:
+        if isinstance(result, pd.Series):
+            return result.to_frame("all")
+        else:
+            return result
+
+
+def set_population_metrics(
```
leewujung, line 186: Similar to the above, I suggest renaming this to `compute_population_metrics`.

echopop/nwfsc_feat/get_proportions.py
```diff
@@ -1,162 +1,1529 @@
...
+    exclude_filter : Dict[str, Any], default {}
+        Groups to exclude, e.g., {"length_bin": [small_lengths]}
+    include_filter : Dict[str, Any], default {}
+        Groups to include, e.g., {"age_bin": [1], "sex": ["female"]}
+
+    Returns
+    -------
+    Union[pd.Series, pd.DataFrame]
+        - pd.Series: if single stratification dimension, returns proportions by strata
+        - pd.DataFrame: if multiple dimensions, returns pivot table with preserved structure
```
leewujung, line 1293: I suggest always returning a dataframe. It is often hard to deal with inconsistent outputs from the same function.

echopop/nwfsc_feat/get_proportions.py
```diff
@@ -1,162 +1,1529 @@
...
+        DataFrame with number proportions containing:
+        - proportion: number proportions
+        - length_bin: length bins
+        - Additional grouping columns (age_bin, sex, etc.)
+    stratify_by : List[str]
+        Columns for stratification
+    exclude_filter : Dict[str, Any], default {}
+        Groups to exclude, e.g., {"length_bin": [small_lengths]}
+    include_filter : Dict[str, Any], default {}
+        Groups to include, e.g., {"age_bin": [1], "sex": ["female"]}
```
leewujung, line 1287: For the 3 `get_*_proportions_slice` functions, only 1 has both `include_filter` and `exclude_filter`, and the other 2 have only `include_filter`. I think that's confusing - better to have them all support both, so user expectation is consistent.

echopop/nwfsc_feat/get_proportions.py
```diff
@@ -1,162 +1,1529 @@
...
+def get_number_proportions_slice(
+    number_proportions: pd.DataFrame,
+    stratify_by: List[str],
+    exclude_filter: Dict[str, Any] = {},
+    include_filter: Dict[str, Any] = {},
+) -> Union[pd.Series, pd.DataFrame]:
     """
-    ds_proportions: xr.Dataset
-    return ds_proportions
+    Extract number proportions for a specific population slice with flexible filtering.
```
leewujung, line 1268: Making the description consistent with the others.
```suggestion
    Extract number proportions for a specific slice of the population.
```

echopop/nwfsc_feat/get_proportions.py
```diff
@@ -1,162 +1,1529 @@
...
+def get_weight_proportions_slice(
+    weight_proportions: pd.DataFrame,
+    stratify_by: List[str],
+    include_filter: Dict[str, Any] = {},
+    number_proportions: Optional[Union[Dict[str, pd.DataFrame], pd.DataFrame]] = None,
+    length_threshold_min: float = 0.0,
+    weight_proportion_threshold: float = 1e-10,
+) -> pd.Series:
+    """
+    Calculate weight proportions for a population slice with optional thresholding.
```
leewujung, line 1385: Making the description consistent.
```suggestion
    Calculate weight proportions for a specific slice of the population with optional thresholding.
```

echopop/nwfsc_feat/get_proportions.py
```diff
@@ -1,162 +1,1529 @@
...
+    -----
+    Thresholding Logic:
+    1. Calculate target group weight proportions normally
+    2. If number_proportions provided, calculate corresponding number thresholds
+    3. Apply threshold mask where both weight and number proportions are very small
+    4. Set masked values to 0.0 to indicate unreliable estimates
+
+    The function handles dictionary inputs by computing element-wise products
+    across all datasets to identify cases where multiple sources agree on
+    very low proportions.
```
leewujung, line 1425: Not sure which are the "dictionary inputs" you refer to here. What are the multiple sources? and which proportion are you talking about here when you say "low proportions" ?

I think try to write in a more concrete way, like just write out what will the function do when X source is available, when Y source is available, and when X and Y sources are both available, will be helpful to users.

Since this is very specific to FEATs hake workflow, I wouldn't worry about writing something generic. The operations and inputs/outputs are already very locked in. This is a very different scenario compared to the inversion classes.

leewujung (commented):
@brandynlucca : Thanks the PR! I've comments on various code sections, please take a look. 

My main thought is that since these operations are very specific to FEAT's hake workflow, the function arguments and docstrings don't need to be flexible, because even if you made those flexible, it still won't work with generic data. This is something we discussed before, and something good to think through when designing functions.

brandynlucca:
This is superseded by #370 since that PR includes the changes within this one. Therefore, this PR will be closed.

PR #356, opened 2025-07-10
Refactored empirical and fitted variogram calculations

brandynlucca:
This PR includes functions that achieve the following: 

- Read in the isobath reference dataframe
- (Optionally) Standardize the coordinates of the transect DataFrame
- Calculate the empirical variogram (with internal functions that calculate things like the distance matrix, azimuth angle filtering, etc)
- Fitting the theoretical variogram based on user-supplied parameters and models. This included making the new refactored code compatible with the previous `variogram` and respective model-specific functions that were already pretty lean

In addition to these changes, new `pytest` functions and fixtures for both the refactored code as well as the previous variogram models (and `variogram` wrapper function). 

One component excluded from this PR is a class that encapsulates a lot of this functionality. It seems to make more sense to leave this to *after* the kriging components are finalized since it makes sense to have a shared class that provides all of the geostatistical methods (i.e. coordinate-system manipulation, variograms, kriging).


echopop/nwfsc_feat/spatial.py
```diff
@@ -0,0 +1,639 @@
...
+from scipy import interpolate
+
+from ..spatial.variogram import variogram
+
+
+def standardize_coordinates(
+    data_df: pd.DataFrame,
+    longitude_offset: float = 0.0,
+    latitude_offset: float = 0.0,
+    reference_df: Optional[pd.DataFrame] = None,
```
leewujung, line 15: Suggest changing to `df_data` and `df_reference` to be consistent with other variable naming.

echopop/nwfsc_feat/spatial.py
```diff
@@ -0,0 +1,639 @@
...
+    Returns
+    -------
+    pd.DataFrame
+        DataFrame with the new standardized coordinates 'x' and 'y'.
+    float or None
+        Total longitudinal distance (degrees) used for standardizing coordinates that can be used
+        for the transformation of other georeferenced datasets.
+    float or None
+        Total latitudinal distance (degrees) used for standardizing coordinates that can be used
+        for the transformation of other georeferenced datasets.
```
leewujung, line 47: The docstrings here reads messy - are you talking about the transformation or standardization step?

echopop/nwfsc_feat/spatial.py
```diff
@@ -0,0 +1,639 @@
...
+    return gamma_h, mean_lag_covariance
+
+
+def empirical_variogram(
+    transect_df: pd.DataFrame,
+    n_lags: int,
+    lag_resolution: float,
+    azimuth_filter: bool,
+    azimuth_angle_threshold: float,
+    variable: str = "biomass_density",
```
leewujung, line 411: `quantity` may be more suitable than `variable`?

leewujung (approved):
@brandynlucca : Thanks for the PR. I have pretty minor suggestions. The only one question below:
- why is there a commented out `initialize_variogram_parameters` function and another active function with an identical name? Remove the commented out version?

Other than that, I started https://github.com/OSOceanAcoustics/echopop/issues/372 to keep track of subpackages that are from before that will continue to be used in the refactored code, for later reorganization/renaming consideration.

Feel free to merge once my comments are resolved.

leewujung:
I think you can incorporate my comments here in #360 as part of the suggested changes (which breaks up the class in a specific way), and then close this PR.

brandynlucca:
I have incorporated and committed the feedback from this PR into the branch associated with #360. This PR is therefore superseded by #360 and can be closed.

PR #360, opened 2025-07-16
Refactored geostatistics class

brandynlucca:
This PR adds the `Geostats` class that encapsulates the coordinate transformation, mesh cropping, empirical variogram calculation, theoretical variogram model fitting, and kriging analysis steps. 

This also includes all associated changes to the associated function docstrings, test functions, and fixtures.

### DIFF
--- a/echopop/nwfsc_feat/kriging.py
+++ b/echopop/nwfsc_feat/kriging.py
@@ -1052,6 +1052,9 @@ class Kriging:
         variogram_params: Dict[str, Any],
         coordinate_names: Tuple[str, str] = ("x", "y"),
     ):
+        """
+        Create class object. If input validation fails, then the object will not be created.
+        """
 
         # Validate
         try:


leewujung (commented):
@brandynlucca : Thanks for the PR. I have reviewed this one and the proceeding ones before you wrapped the steps into a `Geostats` class.

I think it would be better to not wrap things into a single class, because it makes the workflow very locked in. I suggest the following:
- Overall: make the steps being 1) variogram fitting, and 2) kriging
  - this will help with easily plugging in variogram fitting GUI stuff later
- Move all functions related to variogram fitting in `nwfsc_feat/spatial.py` under `nwfsc_feat/variogram.py`.
- Move/rename `echopop.spatial.variogram.py` to `nwfsc_feat/variogram_models.py`.
- Move all functions related to kriging in `nwfsc_feat/spatial.py` under `nwfsc_feat/kriging.py`
- Make `Geostats.git_variogam_model` an independent function under `nwfsc_feat/variogram_models.py`
- Separate out `geo.project_coordinates` to an independent function in `nwfsc_feat/kriging.py`
  - I still prefer `latlon2xy` or `latlon_to_xy` so it is clear what the function is doing
  - I think it's better for people to apply these directly on transect and mesh dataframes (i.e. not hiding what the transformation is applies on)
- I would suggest to go with something similar to what I had in the refactor design, to have a small kriging class that handles mesh creation (`kriging.crop_mesh()`) and kriging (`kriging.krige()`) only
  - `crop_mesh()` will crop the mesh based on a mesh_template
  - have an argument in `crop_mesh()` on how to crop: `transect_ends_crop` and `convex_hull`, accompanied by another argument that contains the parameters in a dictionary, and uses the right cropping method under the hood. I am suggesting this because I feel people shouldn't have to look into the functions in `mesh.py`.
  - `krige()` will perform kriging and return the results in a way similar to `geo.krige()`
  - the `boundary_search_strategy` seems very specific to FEAT data, how about just implementing an argument to `krige()`:  `adaptive_search_strategy=uniform_search`, `adaptive_search_strategy=western_boundary`, or a custom strategy

PR #351, opened 2025-06-14
Refactoring initial transect data processing

brandynlucca:
This includes refactored changes to the steps used for calculating different quantities for the transect data through weight proportions [the last component is still being worked on].

PR #361, opened 2025-07-23
Post-kriging apportionment refactoring

brandynlucca:
This includes the post-processing performed on the kriged mesh population estimates, which achieves the following steps:

* Convert mesh biomass to NASC
* Distribute the kriged estimates over a grid (e.g. age-vs-length for each sex)
* Standardize the unaged distributions to be distributed across age as well with optional imputation
* Combine the aged and unaged tables into one single table
* Redistribute values in the distribution based on some defined criteria, such as excluding age-1 fish

This PR also includes a function for removing specimen-specific haul numbers from the catch data. This avoids incidentally double-counting haul weights where an entire haul's catch was represented in the specimen station dataset and not binned length station data.

All functions are accompanied by corresponding `pytest` functions and fixtures. `coverage` clocks in with 92%, with all misses pointing to interspersed validation lines that raise Errors (e.g. missing columns) that were retained from earlier debugging.


echopop/workflow/feat_hake.py
```diff
@@ -757,4 +757,112 @@
...
+    population_table=dict_kriged_biomass_table,
+    table_names=["aged", "standardized_unaged"],
+    table_index=["length_bin"],
+    table_columns=["age_bin", "sex"],
+)
+
+# ##################################################################################################
+# Redistribute the kriged abundance estimates
+# -------------------------------------------
+# THIS NEEDS TO BE FULLY IMPLEMENTED WITH TESTS, ETC.
```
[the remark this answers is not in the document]
brandynlucca: Ah yep, that is correct.

echopop/nwfsc_feat/apportion.py
```diff
@@ -1,123 +1,704 @@
...
+
+    # Compute biomass if not already in DataFrame
+    if "biomass" not in mesh_data_df.columns:
+        mesh_data_df["biomass"] = mesh_data_df["biomass_density"] * mesh_data_df["area"]
+
+    # Set the link column names
+    mesh_data_df.rename(columns=mesh_biodata_link, inplace=True)
+
+    # Map the link columns
+    mesh_index = [v for v in mesh_biodata_link.values() if v in mesh_data_df.columns]
```
leewujung, line 111: If you just renamed the column, why would the renamed column not exist in the dataframe and needed to be checked?

echopop/nwfsc_feat/apportion.py
```diff
@@ -1,123 +1,704 @@
...
+    missing_mesh_columns = set(list(mesh_biodata_link.keys())).difference(mesh_data_df.columns)
+    if len(missing_mesh_columns) > 0:
+        raise KeyError(
+            f"The following link columns were missing from the mesh DataFrame: "
+            f"{', '.join(col for col in missing_mesh_columns)}."
+        )
+
+    # Compute biomass if not already in DataFrame
+    if "biomass" not in mesh_data_df.columns:
+        mesh_data_df["biomass"] = mesh_data_df["biomass_density"] * mesh_data_df["area"]
```
leewujung, line 105: In the workflow, what are the cases when there is `biomass_density` but not `biomass`?

I think it would be better to just error out people if `biomass` is not already in the dataframe, instead of calculating it under the hood. Hiding this type of calculation within a function makes it harder to trace where different variables are produced.

echopop/nwfsc_feat/apportion.py
```diff
@@ -1,123 +1,704 @@
...
+    ----------
+    reference_table_df : pd.DataFrame
+        Reference table used to determine which values should be imputed and to calculate
+        imputation ratios. Typically contains aged fish data with complete age structure.
+    initial_table_df : pd.DataFrame
+        Original population table before standardization. Used to determine which cells
+        have actual data vs. missing values.
+    standardized_table_df : pd.DataFrame
+        Standardized population table that needs imputation. This will be modified with
+        imputed values.
```
leewujung, line 307: What do you mean by "standardized" here? Do you mean the total sum to 1 along one or more of the dimensions?

echopop/workflow/feat_hake.py
```diff
@@ -757,4 +757,112 @@
...
+# ##################################################################################################
+# Standardize the unaged abundance estimates to be distributed over age
+# ---------------------------------------------------------------------
+
+dict_kriged_abundance_table["standardized_unaged"] = apportion.standardize_kriged_estimates(
+    population_table=dict_kriged_abundance_table["unaged"],
+    reference_table=dict_kriged_abundance_table["aged"],
+    group_by=["sex"],
+    impute=False,    
+)
```
leewujung, line 810: I feel this function (`standardize_kriged_estimates`) can be renamed to something like `distribute_unaged_from_aged`. Even though the underlying operations are general, if I understand this correctly, the only usecase is to distribute unaged biomass or abundance across all ages according to the previously calculated aged proportions.

echopop/workflow/feat_hake.py
```diff
@@ -757,4 +757,112 @@
...
+# ##################################################################################################
+# Consolidate the kriged abundance estimates into a single DataFrame table
+# ------------------------------------------------------------------------
+
+df_kriged_abundance_table = apportion.combine_population_tables(
+    population_table=dict_kriged_abundance_table,
+    table_names=["aged", "standardized_unaged"],
+    table_index=["length_bin"],
+    table_columns=["age_bin", "sex"],
+)
```
leewujung, line 834: Similar to the above, I think this `combine_population_tables` function can be renamed to something like `combine_aged_unaged_kriged_estimates` to be more clear what the underlying operations are.

echopop/workflow/feat_hake.py
leewujung: I also feel `standardized_unaged` could be renamed to something like `unaged_distributed_across_age` to capture the quantity

echopop/workflow/feat_hake.py
```diff
@@ -757,4 +757,112 @@
...
+# Redistribute the kriged abundance estimates
+# -------------------------------------------
+# THIS NEEDS TO BE FULLY IMPLEMENTED WITH TESTS, ETC.
+
+# Re-allocate the age-1 abundance estimates 
+df_kriged_abundance_table_noage1 = apportion.redistribute_population_table(
+    population_table=df_kriged_abundance_table,
+    exclusion_filter={"age_bin": [1]},
+    group_by=["sex"],
+)
```
leewujung, line 857: how about `redistribute_excluded_estimates`?

leewujung (commented):
@brandynlucca : Thanks for the PR. I added a few comments. I read through the code for some of the functions, but not all. 

Two of my comments are about the function names - IMHO these two functions have limited use case and can be renamed to reflect the operations they are designed to do: 
- `standardize_kriged_estimates` distributes unaged estimates to aged estimates according to previously calculated aged proportions
- `combine_population_tables`: combined aged and distributed unaged and already aged estimates.

I've also made notes in #369 re changing these to use xarray dataset/data variables operations.

Other than these and my other small questions, I think this PR is ready to be merged.

echopop/workflow/feat_hake.py
brandynlucca: Another thought on the naming: we had swapped `standardize` with `scale` for the earlier proportion functions in #351.

echopop/workflow/feat_hake.py
brandynlucca: I had somewhat envisioned this to be a bit more generalized beyond just `aged-unaged` since the underlying operation really just aligns any number of population tables to shared indices and columns, and then sums them into a single table. So while the current usage in `feat_hake.py` is `["aged", "standardized_unaged"]`, this could also be as easily applied for `["ship1", "ship2"]`, `["genusA_speciesA", "genusA_speciesB"]`, `["year1", "year2", "year3"]`, etc. But I think a function name like `sum_population_tables`, `merge_population_tables`, or `aggregate_population_tables` are more descriptive descriptive names that better represent the intent of this operation.

echopop/workflow/feat_hake.py
brandynlucca: Add validation error checks -- only return the updated columns rather than partial column coverage.

PR #371, opened 2025-08-06
AFSC to FEAT NASC data format conversion

brandynlucca:
This PR includes an updated function that converts the pre-consolidated AFSC NASC files (via `ingest_nasc.read_nac_file`) to the expected format for the FEAT workflow in `feat_hake.py`. 

This also includes a bug fix to `ingest_nasc.filter_transect_intervals` that did not account for some of the differences between the FEAT and AFSC formats.

This results in the following chain of functions:

`read_nasc_file` -> `filter_transect_intervals` -> `convert_asfc_nasc_to_feat`

The rest of the downstream workflow can therefore be left unchanged when running things for the 1995/1998/2001 NASC datasets.


leewujung (approved):
Question: in the PR description you said:

> This results in the following chain of functions:
> `read_nasc_file` -> `filter_transect_intervals` -> `convert_asfc_nasc_to_feat`

but in the code (`feat_hake.py`), the sequence is:
> `read_nasc_file` -> `convert_asfc_nasc_to_feat` -> `filter_transect_intervals`

I assume the latter (in the code) is the right sequence, but thought to double check.

Other than that, I think this PR is ready to be merged.

PR #376, opened 2025-08-19
Refactor stratified analysis

brandynlucca:
This PR includes:

- Adding `awkward` as a dependency [[f8a04f4](https://github.com/OSOceanAcoustics/echopop/commit/e2a4d97d05d14ba3d970883d7345c4f6ca95dd15)]
- Adding a utility rounding function to get around the default banker's rounding that `numpy` uses. This was causing minor headaches with disagreements over the number of resampling transects per stratum when compared to `EchoPro` [[e284d97](https://github.com/OSOceanAcoustics/echopop/commit/dd57ce28fb0db7959550474570040c4fd909581d)]
- Revamping functions and algorithms for different confidence interval methods, and how they interface with the stratified analysis. This includes updated docstrings [[dd56ce2](https://github.com/OSOceanAcoustics/echopop/commit/338d6dcf7d3d754e40a93df8a2b45c3b85e67dd7)]
- Fixtures and tests for the revamped confidence interval functions [[338d6dc](https://github.com/OSOceanAcoustics/echopop/commit/24a1f2721b5ce4058502e3afcc04dd5799e3ae05)]
- New methods and approaches to running the stratified analysis for transect and kriged data. This includes updated docstrings [[24a1f27](https://github.com/OSOceanAcoustics/echopop/commit/c33c813c4d01d13d5da8dacb401f1346e3854f5c)]
- Fixtures and tests for the stratification analysis [[c33c813](https://github.com/OSOceanAcoustics/echopop/commit/1bd24cc02d78296e6e234399629f7402675ee848)]
- Incorporating the stratified analysis into the `feat_hake.py` workflow file [[1bd24cc](https://github.com/OSOceanAcoustics/echopop/commit/ab48e63f8055ff71c38a679d7772de7f6f875753)]


echopop/nwfsc_feat/stratified.py
```diff
@@ -0,0 +1,1002 @@
...
+        Raises
+        ------
+        RuntimeError
+            If stratified_bootstrap() has not been run prior to calling this method.
+
+        Notes
+        -----
+        The method computes area-weighted survey statistics and includes coefficient
+        of variation estimates. The t-jackknife method is recommended as it provides
+        better coverage properties for small sample sizes common in survey data.
```
leewujung, line 923: The first sentence seems would be better absorbed into the docstring of the returned dataframe. The second sentence can just stay as is.

echopop/nwfsc_feat/utils.py
```diff
@@ -666,3 +721,10 @@ class TSLRegressionParameters(InputModel, title="TS-length regression parameters
 
     slope: float = Field(allow_inf_nan=False)
     intercept: float = Field(allow_inf_nan=False)
+
+
+def roundn(n):
+    """
+    Rounding function to avoid bankers rounding
+    """
+    return n.apply(lambda x: np.floor(x + 0.5) if x >= 0 else np.ceil(x - 0.5))
```
leewujung, line 730: What is the net effect you want to achieve here? The condition tests for if x>=0, so it does not seem to be a general rounding function but something that specifically handles values above/below 0. Depending on the goal, maybe change the function name?

echopop/nwfsc_feat/stratified.py
```diff
@@ -0,0 +1,1002 @@
...
+        self._summarize_survey()
+
+        # Calculate degrees of freedom for variance calculations
+        sample_dof = self._dof()
+
+        # Generate bootstrap sample arrays
+        sampled_distances, sampled_areas, sampled_values = self._prepare_bootstrap_arrays()
+
+        # Compute distance-based stratified weights, handling division by zero
+        mean_distances = awk.mean(sampled_distances, axis=-1, keepdims=True)
```
leewujung, line 836: Is the use of awkward arrays here (and in other places) to avoid having to do list comprehension (loop)? Am trying to understand why, since the calculated results are then converted back to numpy and eventually stored in a dataframe.

echopop/nwfsc_feat/stratified.py
```diff
@@ -0,0 +1,1002 @@
...
+
+        # Initialize the random number generator
+        self.rng = np.random.default_rng(resample_seed)
+
+        # Initialize attributes
+        self.resampled_df = None
+        self.variable = None
+        self.transect_summary = None
+        self.strata_summary = None
+        self.survey_summary = None
```
leewujung, line 94: Could you add in comment here what each attribute stores? `resampled_df` is the main one that triggers this request - I feel there's a more suitable name given the content (so similarly I feel there is probably another name for `_prepare_resampled_dataframe`).

leewujung: maybe something like:
- `self.resampled_df` -> `self.bootstrap_results`
- `self. _prepare_resampled_dataframe` -> `self._format_boostrap_results`

echopop/workflow/feat_hake.py
```diff
@@ -575,101 +603,403 @@
...
+# Compute coefficient of variation (CV) and other estimators for transect biomass
+# -------------------------------------------------------------------------------
+
+# Run bootstrapping procedure
+jh.stratified_bootstrap(data_df=df_nasc_no_age1, 
+                        stratify_by=["geostratum_inpfc"], 
+                        variable="biomass")
+
+# Compute summary statistics for each stratum and overall survey
+transect_results = jh.summarize(ci_percentile=0.95, ci_method="t-jackknife")
```
leewujung, line 971: I think `.summarize` here is a good name, but the operations are quite different from the 3 `._summarize_*` methods. How about changing `._summarize_*` to something more verbose like `._compute_*_statistics`?

leewujung (commented):
@brandynlucca : I like the function breakdowns in this PR. My comments are pretty minor about method/function renaming. Other than those, I think this PR is ready to be merged.

echopop/nwfsc_feat/stratified.py
brandynlucca: By this do you mean moving this text from the `Notes` to `Returns` section?

PR #377, opened 2025-09-01
Refactor matrix inversion

brandynlucca:
This PR includes updated functions, methods, etc. for running the matrix inversion analysis (for both krill and in general). 

**Note**: the tests have pretty spotty coverage at the moment so I am still actively working on these to make sure at least the primary usage is covered and working from there. 

```
Name                                                       Stmts   Miss  Cover
------------------------------------------------------------------------------
echopop\inversion\inversion_matrix.py                        192    117    39%
echopop\inversion\operations.py                               65     35    46%
echopop\inversion\scattering_models.py                        95     82    14%
```


echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+        raise FileNotFoundError(
+            f"The export file directory ({sv_path.as_posix()}) contains no files!"
+        )
+
+    # Update the units for `center_frequencies` to match expected values from Echoview
+    # ---- Hz -> kHz
+    if center_frequencies is not None:
+        center_frequencies = {freq * 1e-3: value for freq, value in center_frequencies.items()}
+
+    # Get the target inversion files
```
leewujung, line 659:
```suggestion
    # Get the target Sv files
```

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+    ingest_nasc.sort_echoview_export_df(sv, inplace=True)
+
+    # Keep target transmit frequencies
+    if center_frequencies:
+        sv_subset = sv[sv["frequency"].isin(center_frequencies)]
+    else:
+        sv_subset = sv.copy()
+        center_frequencies = {
+            freq: {"min": -999.0, "max": 999.0} for freq in sv_subset["frequency"].unique()
+        }
```
leewujung, line 693:
```suggestion
    # Set min/max threshold for each frequency
    if center_frequencies:
        sv_subset = sv[sv["frequency"].isin(center_frequencies)]
    else:
        sv_subset = sv.copy()
        center_frequencies = {
            freq: {"min": -999.0, "max": 999.0} for freq in sv_subset["frequency"].unique()
        }
```

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+        If aggregation method is not one of the supported options
+
+    Notes
+    -----
+    Processing pipeline:
+
+    1. **Thresholding**: Remove noise and artifacts using Sv thresholds
+    2. **Unit conversion**: Convert Sv from dB to linear units for integration
+    3. **NASC calculation**: Compute Nautical Area Scattering Coefficient if missing
+    4. **Frequency standardization**: Convert frequency units to Hz
```
leewujung, line 515:
```suggestion
    4. **Frequency unit conversion**: Convert frequency units to Hz
```

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+    if "sv_mean" in data.columns:
+        mask = (data["sv_mean"] < min_thresh) | (data["sv_mean"] > max_thresh)
+    else:
+        raise KeyError(
+            "Could not apply Sv thresholds. Column 'sv_mean' could not be found within the "
+            "ingested acoustic DataFrame."
+        )
+
+    # Apply the thresholds
+    data.loc[mask, "sv_mean"] = -999.0
```
leewujung, line 177: Why not just set this to NaN? I don't think `sv_mean` needs to be an integer? the "standard" is just what Echoview uses?

leewujung: if you used NaN, then the line after # Drop empty cells is not needed

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+    # Apply minimum and maximum Sv thresholds
+    sv_thresholded = apply_Sv_thresholds(data, sv_thresholds)
+
+    # Compute the linear volumetric scattering coefficient (sv)
+    sv_thresholded["sv_mean_linear"] = 10.0 ** (sv_thresholded["sv_mean"] / 10.0)
+
+    # Drop empty cells
+    sv_reduced = sv_thresholded.loc[sv_thresholded["sv_mean"] > -999.0]
+
+    # Transform frequency
```
leewujung, line 535:
```suggestion
    # Convert frequency units
```

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+    # Calculate area backscattering coefficient (sa) by integrating sv over thickness
+    sa = sv_linear * thickness_mean  # ∫ sv dz approximated as sv × thickness
+
+    # Convert to NASC using sA = 4π(1852)² × sa
+    nasc = STERADIANS_SPHERE * METERS_PER_NMILE_SQUARED * sa
+
+    return nasc
+
+
+def aggregate_cells(data):
```
leewujung, line 236: If I read this function correctly, "aggregate" is not what this function is doing. The function "organizes" the  Sv and NASC data into different entries of a pivot table.

I suggest changing the function to avoid confusion. 

I also added #398 as a reminder to upgrade to use xarray as the data container.

echopop/nwfsc_feat/ingest_sv.py
```diff
@@ -0,0 +1,701 @@
...
+    ]
+
+    # 'Interval' must be present
+    if "interval" not in valid_idx_cols:
+        raise KeyError(
+            "Integration over intervals requires column 'interval' to be present within the "
+            "ingested acoustic DataFrame."
+        )
+
+    # Weight the linear Sv values by thickness
```
leewujung, line 338:
```suggestion
    # Weight the linear sv values by thickness
```

echopop/typing/inversion.py
```diff
@@ -0,0 +1,682 @@
...
+            params.add(
+                name,
+                value=kwargs["value"],
+                min=kwargs.get("min", 0.0),
+                max=kwargs.get("max", np.inf),
+                vary=kwargs.get("vary", True),
+            )
+        return params
+
+    def unscale_dict(self, scaled_dict: dict) -> dict:
```
leewujung, line 324: I think it is confusing to have a `unscale_dict` when there is already `unscale`. The number of use cases for `unscale_dict` is small, so I suggest using `unscale` to perform the operations.

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+    return best_fit_set
+
+
+class InversionMatrix(InversionBase):
+    """
+    Matrix-based acoustic scattering parameter inversion for marine organisms.
+
+    This class performs acoustic inversion to estimate biological parameters
+    (size, density, abundance) from multi-frequency volume backscattering
+    strength measurements. It uses advanced scattering models and nonlinear
```
leewujung, line 534:
```suggestion
    strength measurements. It uses physics-based scattering models and nonlinear
```

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+            Whether to print progress messages during initialization and inversion
+
+        Notes
+        -----
+        The constructor performs minimal initialization, deferring heavy
+        computation until build_scattering_model() is called. This allows
+        for efficient object creation and parameter validation.
+
+        The random number generator is configured based on simulation_settings
+        to ensure reproducible Monte Carlo sampling when mc_seed is provided.
```
leewujung, line 692:
```suggestion
        The random number generator is configured based on simulation_settings.
        To ensure reproducible Monte Carlo sampling, set mc_seed in simulation_settings.
```

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+            - scale_parameters: Whether to scale parameters to [0,1]
+            - minimum_frequency_count: Min frequencies required for inversion
+        verbose : bool, default=True
+            Whether to print progress messages during initialization and inversion
+
+        Notes
+        -----
+        The constructor performs minimal initialization, deferring heavy
+        computation until build_scattering_model() is called. This allows
+        for efficient object creation and parameter validation.
```
leewujung, line 689: I don't think the second sentence adds information.
```suggestion
        The constructor performs minimal initialization, deferring heavy
        computation until build_scattering_model() is called.
```

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+            # ---- Check
+            valid_args = ValidateBuildModelArgs.create(
+                **dict(model_parameters=model_parameters, model_settings=model_settings)
+            )
+        # Break creation
+        except (ValidationError, Exception) as e:
+            raise EchopopValidationError(str(e)) from None
+
+        # Update attributes
+        self.model_params = valid_args["model_parameters"]
```
leewujung, line 862: why do you need to validate the model parameters again when `model_parameters` are already validated from `InvParameters`. I think it is better to just keep it there to not meddle the water here.

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+        self.model_settings.update(valid_args["model_settings"])
+
+        # Update the model settings to include the model function
+        self.model_settings["model_function"] = SCATTERING_MODEL_PARAMETERS.get(
+            self.model_settings["type"]
+        )["function"]
+
+        # Scale the parameters, if defined
+        if self.simulation_settings["scale_parameters"] and not self.model_params.is_scaled:
+            self.model_params.scale()
```
leewujung, line 872: I think having the scaled=True/False both in `simulation_settings` and `model_params` and have one controls the other is messy. Suggest having this flag only in `model_params`, and as suggested in a previous comment, keep both the scaled and unscaled version of the parameters.

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+from ..inversion.inversion_base import InversionBase
+from ..typing import InvParameters, MCInvParameters
+from ..validators.inversion import (
+    ValidateBuildModelArgs,
+    ValidateInversionMatrix,
+)
+from ..validators.scattering_models import SCATTERING_MODEL_PARAMETERS
+
+
+def echopop_optim_cb(params, iter, resid, *args, **kwargs):
```
leewujung, line 20: suggest function name change to make it clear what the function does: `minimizer_print_callback`.

echopop/workflow/feat_krill.py
```diff
@@ -0,0 +1,141 @@
...
+# ==================================================================================================
+
+    scattering_parameters = InvParameters(MODEL_PARAMETERS)
+
+# ==================================================================================================
+# ==================================================================================================
+# INITIALIZE
+# ==================================================================================================
+
+    INVERSION = InversionMatrix(sv_data_sub, SIMULATION_SETTINGS)
```
leewujung, line 117: I would reserve all cap for constants.
```suggestion
    invert_krill = InversionMatrix(sv_data_sub, SIMULATION_SETTINGS)
```

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+        center_frequencies=center_frequencies,
+        **parameter_set,
+        **model_settings["environment"],
+        **model_settings,
+    )
+
+    return objective(Sv_prediction, Sv_measured)
+
+
+def objective(Sv_prediction: np.ndarray[float], Sv_measured: np.ndarray[float]):
```
leewujung, line 226: I suggest absorbing `objective` into `fit_Sv` so that there is one less thing to keep track of. The minimizer is functioning on `fit_Sv` so it would be fine to have that in there. In fact, i'd even suggest renaming it to be something like `cost_Sv` or `objective_Sv_diff`

echopop/workflow/feat_krill.py
```diff
@@ -0,0 +1,141 @@
...
+    "density_sw": 1026.9,    
+}
+
+SIMULATION_SETTINGS = {
+    "environment": ENVIRONMENT,
+    "monte_carlo": True,
+    "mc_realizations": 2,
+    "scale_parameters": True, 
+    "minimum_frequency_count": 2,
+    "reference_frequency": 120e3,
```
leewujung, line 80: This is not used in the inversion so shouldn't be included here.

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+            parameters.reset_index()
+            .drop("number_density", axis=1)
+            .set_index(idx)
+            .reindex(reference_nasc.index),
+        ],
+        axis=1,
+    )
+
+
+def invert_transect_number_density(
```
leewujung, line 1083:
```suggestion
def calculate_transect_number_density(
```
since it is not an inversion.

echopop/inversion/inversion_matrix.py
```diff
@@ -0,0 +1,1230 @@
...
+        * reference_df["nasc_weight"]
+        * acoustic_df["thickness_mean"]
+        * interval_counts
+    ).fillna(0.0) * 1852**2
+
+    # Return
+    return areal_number_density
+
+
+def invert_intervals_number_density(
```
leewujung, line 1160:
```suggestion
def calculate_intervals_number_density(
```
since it is not an inversion.

leewujung (changes requested):
@brandynlucca : thanks for the PR. I read through the code, and had a couple inline comments and have a few questions/comments below:

Right now missing `sv_mean` values are set to -999. Seems better to use NaN for missing `sv_mean` entries.

`echopop/typing/inversion.py` should be moved to under `echopop/inversion/inversion_params.py`.

The use of scaled or original parameters are pretty messy in the class. I suggest just having both (`self.parameters_original` and `self.parameters_scaled` so that there is no need to convert back and forth.

I think `model_parameters` and `model_settings` can be merged into a single dict. The parameters only make sense for the type and other values set in `model_settings`, so the distinction is not clear.

Suggest renaming `scattering_models.py` to `pcdwba.py` since that's what the code includes. We can worry about other models later.

How about moving `estimate_population` and the functions it calls to a separate module called `biology_krill.py` under `echopype/nwfsc_feat`? It should not live together with the inversion code.

It is not obvious to me why `MCInvParameters` is a class inherited from `InvParameters`. I feel it can just be a function that generates the parameter combinations based on the params in `InvParameters`, and then another function that does what `to_lmfit_samples` do as a function. This would remove quite a bit of overhead that comes with the class scaffolding that is not actually used in the code.

Did you observe significant benefits to go with JIT compilation for the calculations under `scattering_models.py`?

echopop/nwfsc_feat/ingest_sv.py
brandynlucca: This was intended to match what `EchoPro` does w.r.t handling empty cells (which is also how Echoview handles this case).

leewujung: If that's the case I think it is safe to save things as NaN instead of -999.

echopop/typing/inversion.py
leewujung: I think it is cleaner to have both the scaled and unscaled parameters BOTH as attributes of the InvParameters object. The optimization method can then be set to use either one of them, but it is not something being handled on the InvParameters side.

echopop/inversion/inversion_matrix.py
brandynlucca: The main reason `model_settings` and `model_parameters` are differentiated is that the values within `model_parameters` are those that are passed into `lmfit.Parameters`, so that requires a format such as: 

```python
{
    "number_density": {"value": 500., "min": 10., "max": 10000., "vary": True},
    "theta_mean": {"value": 10., "min": 0., "max": 90., "vary": True},
    "theta_sd": {"value": 20., "min": 0., "max": 90., "vary": False},
    "length_mean": {"value": 0.030, "min": 0.008, "max": 0.040, "vary": True},
    "length_sd_norm": {"value": 0.15, "min": 0.05, "max": 0.15, "vary": False},
    "g": {"value": 1.015, "min": 1.015, "max": 1.060, "vary": False},
    "h": {"value": 1.020, "min": 1.015, "max": 1.060, "vary":False},
    "radius_of_curvature_ratio": {"value": 3.0, "min": 0.5, "max": 100.0, "vary": False},
    "length_radius_ratio": {"value": 18.2, "min": 14.0, "max": 20.0, "vary": False},
}
```

Since this is passed into `InvParameters`, this would change from:

```python
scattering_parameters = InvParameters(MODEL_PARAMETERS)
inversion_krill = InversionMatrix(sv_data_sub, SIMULATION_SETTINGS)
inversion_krill.build_scattering_model(scattering_parameters, MODEL_SETTINGS)
```

to:

```python
scattering_parameters = InvParameters(MODEL_PARAMETERS)
MODEL_SETTINGS["scattering_parameters"] = scattering_parameters
inversion_krill = InversionMatrix(sv_data_sub, SIMULATION_SETTINGS)
inversion_krill.build_scattering_model(MODEL_SETTINGS)
```

which doesn't seem that different. Perhaps these could be renamed to better describe what they represent? 

I'm not really sure how else to break down the sheer number of parameters for the PCDWBA defined between both `EchoPro`'s `basic_para` and `case_para` sheets from the input `*.xlsx` files (which effectively combine `simulation_settings`, `model_parameters`, and optimization_kwargs`) and `set_para.m` function (which corresponds to the parameters defined in `model_settings`).

PR #381, opened 2025-09-11
Refactor basic plotting functions

brandynlucca:
This PR includes the refactored plotting functions for the transect results, kriged mesh results, and age-length population heatmap. This also includes accompanying tests and fixtures.


brandynlucca:
There is now a `/demo/` folder in `workflows` that includes example outputs based on the new `plotting_demo.py` workflow file. After mulling this PR over a bit, I think it would be more useful in the future to further reduce the plotting functions to just return the `Axes` (or `GeoAxes`) objects and let folks add whatever they want around it. But otherwise, everything has been incorporated so I will merge this into the main branch.

PR #394, opened 2025-09-14
Refactor FEAT report generation

brandynlucca:
This PR includes:
- New function [`distribute_transect_estimates`] that performs a similar operation as `distribute_kriged_estimates`. 
- New class [`FEATReports`] that incorporates a variety of helper functions and methods to produce each report-type 
- Tests and fixtures for all helper functions, `FEATReports` initialization, and methods
- Minor bug fixes


leewujung (changes requested):
@brandynlucca : Thanks for the PR. I have an inline comments about merging `distribute_krige_estimates` and `distribute_transect_estimates`, and some other comments below:

Could you add a notebook to show the exact steps to generate report using the class? I am not able to find it. I added an issue to track that. #397 

`test_FEATReports` sits in `test_class.py` which only has 1 test underneath. Would be better to just change that to `test_FEATReports.py`. Are there other classes you plan to test in this script?

PR #396, opened 2025-09-19
Refactor diagnostic plots

brandynlucca:
This PR includes a refactored version of the diagnostic plots in the form of the `Diagnostics` class. This PR also includes a demo notebook (`workflow/diagnostic_plotting_demo.ipynb`).


echopop/workflow/feat_hake.py
```diff
@@ -996,6 +1002,14 @@
...
     # Biomass table - kriged data
     df_kriged_biomass_table.to_pickle(FILES_DIR / "df_kriged_biomass_table.pkl")
+    # Stratified results - transect data
+    transect_results.to_pickle(FILES_DIR / "stratified_transect_results.pkl")
+    # Stratified results - kriged data
+    kriged_results.to_pickle(FILES_DIR / "stratified_kriged_results.pkl")
+    # Stratified replicates - transect data
+    transect_replicates.to_pickle(FILES_DIR / "stratified_transect_replicates.pkl")
+    # Stratified replicates - kriged data
+    kriged_replicates.to_pickle(FILES_DIR / "stratified_kriged_replicates.pkl")
```
leewujung, line 1012: Suggest adding `jh_` as prefix to the `*_results` and `*_replicates` variables here for clarity.

brandynlucca:
> Since all results under the # Pickle section are dataframes, I think we should save them as csv files instead of pickles so that it is generally usable outside of python. Make sure the datatypes are correct when you read them in.

I suppose this (i.e. L978-1016) wasn't necessarily meant to be a generalized case but rather retaining the same `pandas.DataFrame` indexing across script/notebook files. While a `*.csv` would function fine for dataframes like `df_nasc_no_age1_prt` and `df_kriged_results`, it would be somewhat cumbersome to do for others like `df_kriged_abundance_table`. Consequently, it may make more sense to just _remove_ this code block from `feat_hake.py` since it doesn't actually represent a step intended for the workflow but rather something out of convenience. For the purposes of the plotting demo in the notebook and base plotting, I think there is also something to be said about the utility of just calling in pickled files that are already in an intended state, rather than needing to add a bit of data wrangling at the top of the notebook file.

PR #403, opened 2025-10-01
Year-specific hake survey workflows

brandynlucca:
This PR includes year-specific workflows for all survey years between 1995 and 2023. An overview notebook accompanies the files that points out the parameters editable at the top. I included some quick CLI hooks in the files for ease-of-testing (i.e. making sure each file ran without needing to open each one up), but these are otherwise not actually implemented in any robust manner. 

Importantly, the workflow files represent hake survey processing for age-1+ fish extrapolated over the entire kriging mesh regions where the biodata are stratified by `KS` and the stratified analysis is done using the `INPFC` strata. I was thinking it would be best to specify the age-2+ workflow (and other variants such as cropping) in a separate PR (or PRs) since the files in this PR are the "base" workflows that these special cases build upon.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -21,7 +21,6 @@ traitlets
 ipympl
 jupyter_bokeh>=3.0.5
 # Static visualization stack
-cartopy
 matplotlib
 # Interactive visualization stack
 bokeh


leewujung:
> @leewujung:
> 
> > This notebook should probably moved to the docs?
> 
> I think that makes sense since it is just a general description of the year-specific files. I can just run it locally and keep/store the outputs in the notebook so they render via `jupyter-book`, and that should suffice?

Yep!

PR #405, opened 2025-10-06
Add additional example workflows for hake survey processing

brandynlucca:
This PR includes two new workflows:
- `feat_hake_age1_excluded.py`: Running the complete workflow when age-1 fish are excluded, with additional options for removing age-1 dominated hauls for specific survey years.
- `feat_hake_kriging_without_extrapolation.py`: Running the hake survey workflow through the geostatistics where the kriged estimates are not extrapolated over the entire ingested mesh.

Additions:
- Transect-mesh-region definition generating functions have been created for every year, which were all adapted directly from the associated `EchoPro` `*.m` files in `EchoPro_matlab/input_files/Region_def_files/*`. At least in the version I have, these were only generated for survey years between 1995 and 2019. 

Bug-fixes:
- For cases where age-1 dominated hauls are removed, there were resulting `NaN` values that contaminated pre-kriged (i.e. transect) population estimates. This is due to cases where the removal of particular hauls can encompass one or more strata in their entirety, so the merging/reindexing operations are unable to resolve an appropriate value. This has been rectified.

### DIFF
--- a/echopop/nwfsc_feat/FEAT/parameters.py
+++ b/echopop/nwfsc_feat/FEAT/parameters.py
@@ -85,3 +85,1131 @@ def transect_mesh_region_2019(
         transect_upper_bound = [i + 0.4 for i in range(transect_start, transect_end + 1)]
 
     return transect_start, transect_end, transect_lower_bound, transect_upper_bound
+
+
+def transect_mesh_region_2017(
+    region: np.number,
+) -> Tuple[np.number, np.number, List[np.number], List[np.number]]:
+    """
+    Generate region-specific transect boundaries for the 2017 NWFSC survey.
+
+    This function defines the spatial boundaries for three distinct survey regions used in the
+    2017 Northwest Fisheries Science Center (NWFSC) survey. Each region has specific transect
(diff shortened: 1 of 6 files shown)


leewujung (approved):
Hey @brandynlucca : Thanks for the PR!

Do you think you can highlight in these script where the variations from the default workflow are? The scripts are long so I am thinking this would make it easier for people to find out where things are quickly. This would serve kinda duplicated functions from what you have in the docs, but will likely help with memory consolidation.

For the `transect_mesh_region_*` functions: Please make sure there's a page/section describing this in the RTD so that people understand why these are coded up like functions and not just arguments passed to the same function.

brandynlucca:
@leewujung: 

> Do you think you can highlight in these script where the variations from the default workflow are? The scripts are long so I am thinking this would make it easier for people to find out where things are quickly. This would serve kinda duplicated functions from what you have in the docs, but will likely help with memory consolidation.

What would be the best way to highlight this though (aside from including `# !` or all-caps comments)? Like referencing specific line-numbers in the comments that are at the very top of the script files?

Issue #404, opened 2025-10-01
Unexpected outputs from the off-effort transect filter


brandynlucca:
The `ingest_nasc.filter_transect_intervals` file produces unexpected outputs (when compared to `EchoPro`) for certain years, such as 2005. This may be due to the handling of particular edge cases and missing values. 

- [x] 2012 [_currently okay_]
- [x] 2011 [_currently okay_]
- [x] 2009 
- [x] 2007
- [x] 2005 [_many missing values; largest deviation_]
- [x] 2003

PR #408, opened 2025-10-09
Apply fixes to off-effort transect filtering

brandynlucca:
The off-effort transect filter (`ingest_nasc.filter_transect_intervals`) was not producing the same sized dataframes as expected when comparing to the outputs from `EchoPro`'s `remove_unwated_zeros.m`. This has now been rectified and tested for all years between 2003 and 2012 (as noted in #404).

### DIFF
--- a/echopop/workflows/nwfsc_feat/functions.py
+++ b/echopop/workflows/nwfsc_feat/functions.py
@@ -794,9 +794,6 @@ def filter_transect_intervals(
     ):
         transect_filter_df.rename(columns={"transect": "transect_num"}, inplace=True)
 
-    # Store original columns
-    original_columns = nasc_df.columns.tolist()
-
     # Apply a filter, if needed
     if subset_filter is not None:
         # Extract tokens from string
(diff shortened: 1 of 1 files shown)


echopop/nwfsc_feat/ingest_nasc.py
```diff
@@ -1272,23 +1269,105 @@ def filter_transect_intervals(
...
+        nasc_data_indices = np.where(nasc_transect_nums == transect)[0]
+        # ---- Skip if empty
+        if len(nasc_data_indices) == 0:
+            continue
+        # ---- Initialize current transect removal indices
+        current_removal_indices = []
+        # ---- Iterate through gaps
+        if num_intervals > 1:
+            # ---- Multiple intervals case
+            for j in range(num_intervals):
```
leewujung: Would be a good idea to avoid using `j`? How about just `interval`?

brandynlucca:
@leewujung: 
> I only have a very minor comment above, but I do have a request: Could you can add a schematic that shows the procedure in # Process each unique transect that has filter intervals? Seems like that would be more intuitive (since you already sorted out what Chu's operations were doing). Thanks!

Sure! Do you mean a schematic in the documentation (for the implementation?)? Or in the docstring?

PR #415, opened 2025-10-31
[389] Add support for bio data ingest from postgres db

### LINKED ISSUE #389: Add the functionality to ingest biological data directly from database views
Currently `load_data.load_biological_data` will load a set of CSV files into a dictionary of dataframes, which contain data that will be used to:
- compute mean sigma_bs and average weight
- compute number and weight proportions for reapportioning biomass

https://github.com/OSOceanAcoustics/echopop/blob/82970fb2afeb53daddd3248c302b569a04cc24bd/echopop/workflow/feat_hake.py#L169-L209

This set of CSV files are generated from database queries, so it would be useful to cut out the CSV step and get these data directly via database views. This will make the workflow more streamlined and more robust against arbitrary changes in CSV files.

We should still keep the current CSV biological data loading function, so that we can easily create a comparison function to double check the database view function is producing the same results.

cc @aliciabillings-noaa @ElizabethMPhillips

dbashford-NOAA:
Resolves #389 

- Added a function to load bio data from a postgres db as an alternative to loading from a CSV
- Uses `psycopg` package to connect to postgres


dbashford-NOAA:
@leewujung @brandynlucca It looks like the tests are failing because of the `psycopg` dependency for my changes. 
I've added it to `requirements.txt` but I only see `requirements-dev.txt` in the PR workflow. 
Do I need to add `psycopg` somewhere else for the workflow?

dbashford-NOAA:
@leewujung 
It took some testing to figure out but I got that action to work and fixed the issue.
It's passing all the tests and is ready for review now.

leewujung:
@dbashford-NOAA : Awesome! I am glad it worked out! Looking forward to reviewing this (and all the other reviews I owe you...).

echopop/ingest/biological.py
```diff
@@ -112,6 +113,102 @@ def load_biological_data(
     return biodata_dict
 
 
+def load_db_biological_data(
+    db_credentials: Dict[str, str],
+    biodata_sheet_map: Dict[str, str],
```
leewujung: Looks `biodata_sheet_map` is not used and can be removed

leewujung (changes requested):
Hey @dbashford-NOAA: Thanks for the PR and making postgres tests in CI work!

I think this PR is almost ready to go, just have some consistency issues that I marked in inline comments.

In addition, how about changing the biodata loading function names so that it is clear where the data are from?
- `load_biological_data` --> `load_biological_data_excel`
- `load_db_biological_data` --> `load_biological_data_database`

dbashford-NOAA:
@leewujung 
I've made those changes so it should be ready now!

leewujung (approved):
@dbashford-NOAA : Great, thanks! Everything looks fine. I will merge this after the hackday tomorrow in case of breaking something due to the function name change for the bio data ingestion part.

ElizabethMPhillips:
Following up on this and after checking with Wu-Jung today, @dbashford-NOAA please ensure that the updated materialized views in the nwfsc_public_prod.acoustics schema match the previous views developed, and then go ahead and update the 
code to access the nwfsc_public_prod.acoustics schema. After that is complete, we can test pulling the new, merged length view (that includes both length and specimen in one table)

PR #428, opened 2025-12-25
Xarray migration

brandynlucca:
This is a draft PR for the xarray migration.

PR #432, opened 2026-01-07
Updated documentation for xarray migration

brandynlucca:
This represents the updated documentation for the functions and workflows post-xarray migration.


leewujung (commented):
@brandynlucca : I think you should go ahead to merge this so that at least most of the docs are up to date. I am not able to review this PR due to the extreme time limit. I will write an issue for check all the data types are correct in case of missed errors in #428.

Issue #461, opened 2026-02-18
New functions for weight proportions incorporating net selectivity


brandynlucca:
- [ ] Sketch out alternative workflow/steps that incorporates this for meeting next time

Then...
- [ ] Determine where the net selectivity function should be applied to adjust number proportions 
- [ ] Create weight proportions function that incorporates number proportions instead of binned weights

Issue #429, opened 2026-01-02
Weight proportion calculation after adjusting for net selectivity


leewujung:
There is something that we did not realize when discussing net selectivity last time. It is related to how we should estimate the weight of fish that would have been caught had the net not been selective for only larger fish. Below I will first explain the procedure used in EchoPro/Echopop to compute weight proportion, and then discuss where the problems come in.

### Data
- Station 1: unaged fish, weighted by the entire haul
- Station 2: aged fish, individually weighted

### Current weight proportion estimation procedure
1. Derive the length-weight relationship for each stratum from station 2 data
2. Compute the **weight proportion of aged fish across (sex, length, age)** by:
  a. Compute the overall weight proportion across (sex, length, age) by dividing the station 2 weight data over these dimensions by the per-stratum total weight
3. Compute the **weight proportion of unaged fish  across (sex, length)** by:
  a. Estimate the weights for all length bins using the length distribution and the length-weight relationship
  b. Compute the within-unaged-group weight proportions across (sex, length) using the weights estimated in a.
  c. Scale the within-unaged-group weight proportions across (sex, length) to proportion over all fish (unaged + aged); the scaling factor is the proportion of unaged fish in all fish

### Problem
When we correct for net selectivity, we do the correction by adjusting the length distribution of fish. We would need to make assumptions for all 3 steps above:
1. We would need to assume the length-weight relationship derived from station 2 data holds true for smaller fish that may not have been caught by the net
    - This is likely fine?
2. **[RPOBLEM!]** Here we do not have an easy way to compute the weight proportion of fish that would have been caught without a selective net, because the weight proportions of aged fish were computed using the true weights measured in station 2.
    - Potential solution: We could use the length-weight relationship to "patch in" the fish that would have been caught and use those in this calculation.
3. **[RPOBLEM!]** Here we do not have an easy way to estimate the total weight of small unaged fish that would have been caught (used in step 3c. above)
    - Potential solution: As in 2., we could use the length-weight relationship to "patch in" the fish that would have been caught and use those in this calculation.

### Another potential solution
Another potential solution that I think would resolve the problem once and for all (and remove a lot of downstream post-kriging re-distribution messiness) is to simply use estimated weights from the length-weight regression relationship to compute the weight proportion for unaged, aged, and all fish.

@sabrinabeyerNOAA @ElizabethMPhillips @brandynlucca : Let's discuss during the hack day! But I wanted to write this up so that we have something to reference.

[...]

PR #464, opened 2026-02-21
Add net selectivity corrections

brandynlucca:
This PR adds net selectivity correction functions (related to #429 and #461) via these changes: 

- A new `selectivity` submodule with the functions `correct_number_proportions` and `correct_weight_proportions`. The former ingests an argument, `selectivity_params`, which is a dictionary input that can be parameterized with either: 
   - `slope`/`intercept`: logistic regression coefficients
   - `l50`/`sr`: derived selectivity parameters
- A small tweak to the `compute_abundance` function to account for minor format change in the proportions
- An example workflow, `net_selectivity_workflow.py`, has been created to demonstrate functionality for downstream calculations (through the apportionment of kriged estimates)


leewujung (changes requested):
@brandynlucca : Thanks for changing this PR to reflect our discussion yesterday. I have a few comments below, and one of them is more overarching related to `stratum_dim` that may be better addressed in this other PR to address the variable naming and handling related to stratum.

--------------------------------------------------
Change `stratum_dim` to simply being a str instead of a list
- I suggest this because having stratum_dim being just a str makes it very clear in the under-the-hood code which dimensions the operations are supposed to work on. For example:
   ```python
        weight_prop_sex = number_proportions["proportion"].sum(
            dim=[d for d in number_proportions["proportion"].dims if d not in stratum_dim + ["sex"]]
        )
   ```
   can be changed to 
   ```python
        weight_prop_sex = number_proportions["proportion"].sum(
            dim=set(number_proportions["proportion"].dims).difference(["stratum_ks", "sex"])
        )
   ```
   which is easier to read.
   It is often useful for functions like this to do one thing instead of many things.
- This is more overarching related to `stratum_dim` that may be better addressed in this other PR to address the variable naming and handling related to stratum.

-----------------------------------------------
There seems to be a mixed use of "proportion" and "proportion_overall" -- I am not sure if you are still in the middle of finishing this PR though. As we discussed yesterday, it will be good for the code to just have "proportion", which contains the proportion wrt across sex, length, and age within a single stratum (i.e.:
```python
assert `var["proportion"].sel(stratum_dim=1).sum() == 1  # var can be number proportion or weight proportion
```

-------------------------------------------------
I think it is cleaner to not overload the `fitted_weight_proportions` function -- i.e.:
- have `fitted_weight_proportions` just do what it did before for the aged/unaged separated workflows in all previous years
- have a separate function `weight_proportions_all_samples` to compute the weight proportions for the new workflow in which all aged/unaged samples are combined

Rationale:
- it is difficult for users to remember that a different set of weight proportions are used only when `weight_data` and `reference_weight_proportions` are supplied
- I notice that in your current implementation, the number proportion across length (`number_prop_length`) is not sex-specific. I think we agreed on using sex-specific distributions for number and weight distributions over length bin and age bin. I know Chu did different things in EchoPro, but with the new aged/unaged combined workflow, we should do what is more scientifically sound. So:
   - `number_prop_length` should have dims (stratum, sex, length_bin)
   - `weight_prop_length` should have dims (stratum, sex, length_bin)
   - `weight_prop` should have dims (stratum, sex, length_bin, age_bin)

echopop/workflows/nwfsc_feat/workflows/net_selectivity_workflow.py
```diff
@@ -0,0 +1,642 @@
...
+# ==================================================================================================
+# COMPUTE COUNT DISTRIBUTIONS PER AGE- AND LENGTH-BINS [AT STRATUM-LEVEL]
+da_count_distribution_strata = da_count_distribution_hauls.sum(dim=["uid"])
+
+da_count_distirbution = proportions.compute_binned_counts(
+    data=specimen_data.dropna(subset=["length"]),
+    groupby_cols=["stratum_ks", "length_bin", "age_bin", "sex"],
+    count_col="length",
+    agg_func="size",
+)
```
leewujung: This doesn't seem to be used anywhere? Please remove if that's indeed the case.

PR #477, opened 2026-03-18
[389] Support ingest from postgresql with testing

### LINKED ISSUE #389: Add the functionality to ingest biological data directly from database views

dbashford-NOAA:
Resolves #389

Added a function to load bio data from a postgres db as an alternative to loading from a CSV

Cleaned version of PR #415


leewujung:
@dbashford-NOAA : Agreed! I just closed #415.

leewujung (approved):
Thanks @dbashford-NOAA ! I think the code/tests look good on my side. Just requested @aliciabillings-noaa to review. :)

PR #507, opened 2026-07-01
Lat lon update

rebeccathomas-NOAA:
Creating a pull request to update code dealing with [506](https://github.com/echostack-org/echopop/issues/506) regarding reading in Echoview exports with multiple lat/lon columns.


leewujung:
@rebeccathomas-NOAA : what do you think if we just expose what these strings should be as an argument (default to _e or _s)? then you could use different ones depending on the year.

rebeccathomas-NOAA:
Hi Wu-Jung-
I think that would be fine.  Supposedly most years should use _s, but I am not really sure that always happened.
Thanks,
RT

